PR #16, opened 2025-03-22
Add declarative macro guideline: avoid specializing

PLeVasseur:


PLeVasseur:
Hey @pellico, @vccjgust -- I tried to mock up a coding guideline for declarative macros.

Could y'all take a look when able?

src/coding-guidelines/macros.rst
```diff
@@ -5,3 +5,144 @@
...
+   :recommendation: encouraged
+
+   Description of the guideline goes here.
+
+   .. rationale::
+      :id: rat_zqr9uEqP6nzW
+      :status: draft
+
+      It is common to use declarative macros to implement traits which would
+      otherwise involve repetitive code.
```
jgust, line 24: I would say that a derive macro is the more idiomatic way of generating trait implementations, while a declarative macro is more suitable as a syntax extension mechanism as well as for generating boiler plate code outside of places where derive macros are usable (trait implementations can of course be considered as boiler plate at times). 

Perhaps we should phrase this so that we don't give the impression that generating trait implementations via `macro_rules!` is somehow a preferred or recommended way?

src/coding-guidelines/macros.rst
```diff
@@ -5,3 +5,144 @@
...
+
+      In a declarative macro the ordering of the patterns will be the order that
+      they are matched against which can lead to unexpected behavior in the case
+      where we have unique behavior intended for a particular expression.
+
+      If needing to specialize logic within the macro based on a particular
+      expression's value, it may be better to not use a declarative macro.
+
+      Limitation: Note that following this rule means that we are unable to support
+      variadic declarative macros with one or more arguments.
```
jgust, line 34: Unsure if if I follow, but doesn't the repeat syntax in `macro_rules!` allow for specifying handling for zero or more arguments as one rule? Its been a while since we wrote a macro so I am not 100% sure.

jgust (commented):
Just a few initial comments. I will continue my review tomorrow.  Good work so far.

src/coding-guidelines/macros.rst
PLeVasseur: Good point. I think that a derive macro will often be the more idiomatic thing. There is a bit of trade-off to be had here where that requires bringing in external crates (proc-macro2, syn, quote) which could make safety-qualification more challenging.

I think this gets into the discussion we had in the last coding guidelines meeting that Rust is a sophisticated language with a number of ways to do things.

I could see rephrasing this to the below to avoid biasing one way or another in this guideline:

> It's common to use macros to avoid writing repetitive code, such as trait implementations. It's possible to use derive macros or declarative macros to do so.

I think we could also have another guideline which said just what you pointed out that we'd recommend using a derive macro for repetitive trait implementation, if the additional external crates are not of issue. What do you think?

src/coding-guidelines/macros.rst
PLeVasseur: Oh, you're totally right! Thanks for pointing this out.

I took [this example](https://doc.rust-lang.org/rust-by-example/macros/variadics.html) from Rust by Example and came up with [this example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f606fc667995612b81b429c091135fc3) which shows that it's possible to support variadic declarative macros even with a single macro rule at a certain level of nesting.

Will update.

PLeVasseur: Done!

scrabsha:
Do we want to warn against this as well:

```rs
macro_rules! impl_safety_trait {
    // Special pattern for EmergencyValve
    (EmergencyValve) => {
        impl SafetyCheck for EmergencyValve {
            fn verify(&self) -> SafetyLevel {
                // Emergency valve must be open for safety
                if !self.open {
                    SafetyLevel::Red
                } else {
                    SafetyLevel::Green
                }
            }
        }
    };
    // Generic pattern matches any type - including EmergencyValve
    ($t:ty) => {
        impl SafetyCheck for $t {
            fn verify(&self) -> SafetyLevel {
                SafetyLevel::Green
            }
        }
    };
}
```

(The two rules have been swapped)

Both macros (the one from the PR and mine) use a specialized pattern, but your example suggests the code is bad because the specialized rule is dead code.

Could we either:

- (If we intend to forbid the example from my comment) Add a sentence that explicitly says that this rule applies regardless of the ordering of the rule
- (If we intend to allow the example from my comment) state that this guideline only applies to specialized patterns that are dead code (and maybe change the guideline name)

src/coding-guidelines/macros.rst
PLeVasseur: Thanks for the feedback @manthonyaiello! I feel like we're in one of those "depends on your context" scenarios.

This coding guideline is setting out to suggest something for declarative macros specifically, which would be in libcore.

How do you feel about the text that's there currently more as background in this coding guideline?
> It's common to use macros to avoid writing repetitive code, such as trait implementations. It's possible to use derive macros or declarative macros to do so.

On another note, we could have another coding guideline we can give rationale on the circumstances:
* if you need safety-certification, probably stick to libcore and declarative macros for use cas of DRY
* if you are comfortable without safety-certification, derive macros can offer a bit more ergonomics

manthonyaiello: If the purpose of the coding guideline/standard is overall safety-critical use cases, then I think certification can/should be assumed and we should focus on those needs. I would lean away from recommending alternative approaches until/unless they can be met by a certifiable/certified runtime. Just my 2¢.

PLeVasseur:
> Do we want to warn against this as well:
> ... snip ...
> Both macros (the one from the PR and mine) use a specialized pattern, but your example suggests the code is bad because the specialized rule is dead code.

Good catch!

> Could we either:
> 
> * (If we intend to forbid the example from my comment) Add a sentence that explicitly says that this rule applies regardless of the ordering of the rule
> * (If we intend to allow the example from my comment) state that this guideline only applies to specialized patterns that are dead code (and maybe change the guideline name)

I would rather prevent all specialized patterns for the sake of preventing accidents from occurring when someone is refactoring the code / adding new macro rule and so on. What do y'all think?

(By the way the code example is intended to showcase what can go wrong if the guideline is not followed, not be exhaustive in my opinion. I suppose I could be explicit that it doesn't matter on ordering, it's always disallowed if we think necessary?)

pellico:
I am fine with present rule but I don't feel comfortable to start with a very specific rule with a little bigger context.
I think it would be helpful some direct discussion to align on general concept direction.
I just try to write shortly some general consideration about macros in safety context.
I think that macros can be abused and lead to a code that is difficult to test and review.

### Declarative macro
#### Pro
easy to write, little compilation cost
#### Cons
If they are too complex, it is difficult to review them.
If you use cargo expand,  the generated could could be not compiled.
How to measure test coverage ?
### Procedural macro
#### Pro
Expanded code can be compiled
Easier to measure coverage
Easier to create test for procedural macro itself.
### Cons
As reported by @PLeVasseur and confirmed by a compiler vendor, procedural macro can be an issue in compiler assessment. I think [watt](https://github.com/dtolnay/watt) could be a solution for that.
Slow compilation time. 

Some general ideas for guidelines

### Declaration macros

Declarative macros should be short and easily understandable. --> Create some complexity index, line of code, number of arguments etc...
No recursive macros
Test coverage shall be measured on generated code (is this possible ?)

### Procedural macros

Procedural macro package shall be tested and qualified standalone. If it is not possible to qualify the package for the appropriate level the generated code shall be reviewed and then compiled. 
Generated code of procedural macro shall follow the safety coding guidelines as for human written code. (Rationale: It should be reviewed by a human)
No recursive macro
Test coverage shall be measured on generated code (I think this should be easily achievable)

src/coding-guidelines/macros.rst
jgust: My point was not intended to advocate for one approach over another, but rather that we should avoid giving the impression that a non-idiomatic way is recommended. I know it was not the intention but the original text could maybe have been interpreted that way, was my concern.

We can discuss it at the meeting how to best approach these types of situations where the idiomatic way is perhaps more difficult to certify since that is obviously an important point. But just a note: neither declarative or derive macros impose any additional run-time components on their own since both are compile-time code generation mechanisms. But the latter method adds (in practice) additional "semi-standard" third party crates as Pete pointed out.

PLeVasseur:
> I am fine with present rule but I don't feel comfortable to start with a very specific rule with a little bigger context. I think it would be helpful some direct discussion to align on general concept direction. I just try to write shortly some general consideration about macros in safety context. I think that macros can be abused and lead to a code that is difficult to test and review.
> ... snip ...

These are great considerations! Putting an idea out there -- perhaps @pellico you could "scaffold out" some of these thoughts on the lead-in section of the Macros chapter before we get to the guidelines themselves as a PR we can review. What do you think?

jgust:
> > I am fine with present rule but I don't feel comfortable to start with a very specific rule with a little bigger context. I think it would be helpful some direct discussion to align on general concept direction. I just try to write shortly some general consideration about macros in safety context. I think that macros can be abused and lead to a code that is difficult to test and review.
> > ... snip ...
> 
> These are great considerations! Putting an idea out there -- perhaps @pellico you could "scaffold out" some of these thoughts on the lead-in section of the Macros chapter before we get to the guidelines themselves as a PR we can review. What do you think?

I agree, very good considerations! I think we should also mention (either in the lead-in or even as its own guideline), that since writing macros is a relatively rare event, compared to writing normal Rust code, and it has slightly different syntax it is important to properly document each match pattern, where the expected input and corresponding intended output is clarified. This helps both during review as well when you come back to a macro months later and have already forgetting everything about macro writing :-) What do think?

PLeVasseur:
> One small sentence I would suggest adding at the very top is a quick "Rust macros are not C-like preprocessor macros"
> ... snip ...
> I think those two points are important because a lot of people have "PTSD"-like responses from past C-macro insanity, and we want to emphasize that "Rust macro" and "C macro" are as equivalent as "car" and "carpet".

I think that's fair to add as some preamble ahead of the guidelines proper. I need to read the [The Little Book of Rust Macros](https://veykril.github.io/tlborm/) more to be able to capture the nuance here probably.

I had read for example that declarative macros are _partially_ hygienic recently for example, [link](https://veykril.github.io/tlborm/decl-macros/minutiae/hygiene.html).

Perhaps we should provide links to these sorts of resources in each chapter for more details. What do you y'all think?

pellico:
> @PLeVasseur Very nice - thank you - and the commentary here is great!
> 
> One small sentence I would suggest adding at the very top is a quick "Rust macros are **not** C-like preprocessor macros"
> 
> The salient points are:
> 
> * Rust macros are [hygenic](https://rustc-dev-guide.rust-lang.org/macro-expansion.html#hygiene-and-hierarchies), meaning that Rust's macro system prevents errors due to names introduced within a macro
> * Rust macros operate on streams of [**typed** tokens](https://doc.rust-lang.org/book/ch20-06-macros.html) as part of the AST, as opposed to the **un**-typed token streams that the C preprocessor operates on
> 
> _(Well, one sentence and two bullet points 😆)_
> 
> I think those two points are important because a _lot_ of people have "PTSD"-like responses from past C-macro insanity, and we want to emphasize that "Rust macro" and "C macro" are as equivalent as "car" and "carpet".

Definitely declarative macros are "safer" than C macros but procedural macro has greater power than C macro. It can take a piece of code and transform it in a completely different and surprising code. I think that in functional safety scope the usage of macros shall be restricted a lot.  Doesn't from great power come a great responsibility ?
Moreover I think that there is a tendency to use too much procedural macro see below picture from [lib.rs](https://lib.rs/stats) site.

image

jgust:
> Definitely declarative macros are "safer" than C macros but procedural macro has greater power than C macro. It can take a piece of code and transform it in a completely different and surprising code. Bad procedural macro can create more damage than C macro. I think that in functional safety scope the usage of macros shall be restricted a lot. Doesn't from great power come a great responsibility ?

I agree, it should be used with care. But, it is relatively straight forward (and could probably be improved even further) to see what a given macro generates as its output (i.e. the actual code that goes into the binary) from a given input. I would say the focus should be on coming up with guidelines and procedures that give use maximal insight and control of the code being generated to ensure that it conforms to the same safety/quality goals as the handwritten code.

adfernandes:
@PLeVasseur Nice catch on the _partial_ hygenicity of macros!

@pellico I absolutely understand that macros _can_ be abused, and I've certainly seen instances where they _have_ been abused, but I'm not certain that things like "library percent" are great metrics for "good _vs_ bad".

For example, the std library has a _lot_ of instances of `unsafe`, much more than your "usual" Rust crate, but that's because the std library **encapsulates tricky things so you don't have to deal with it**.

Consider a `proc_macro` like the following: `midpoint!(x, y)`. It computes `(x + y)/2)`, which, seems just fine... except it has a hidden integer overflow.

Or another example: `hypotenuse!(x, y)` computed via `sqrt(x*x + y*y)` for floats, which has overflow and underflow for floats, which can be tricky to find.

Those are "bad" macros because the underlying algorithms are subtly incorrect, and that fact **might** be hidden somewhere in "macro language", not canonical Rust.

What about `bake_a_pizza!()` that _actually_ tries to bake _two_ pizzas and calls `release_the_kraken()`? That's a "bad" macro because it's name is misleading, and it has a surprising, **global**, side-effect. One does not expect to release the Kraken when one wants to simply bake a pizza.

Instead, what about guidelines along the lines of:
* Are the macros clear and unambiguous?
* Are they correctly and thoroughly and **exhaustively** documented?
* Are there side effects? If so, what scope? Local? Crate? "Global"?
* Are they _thoroughly_ tested and validated?

These are **much** easier goals for macros in Rust _vs_ macros in C!

pellico:
@adfernandes 
> I absolutely understand that macros _can_ be abused, and I've certainly seen instances where they _have_ been abused, but I'm not certain that things like "library percent" are great metrics for "good _vs_ bad".

I agree that the "library percent" is not a metric for  "good _vs_ bad". I just want to highlight that I have perception that devs in Rust use more macros than in C probably because they are more powerful.

src/coding-guidelines/macros.rst
```diff
@@ -5,3 +5,142 @@
...
+
+   Description of the guideline goes here.
+
+   .. rationale::
+      :id: rat_zqr9uEqP6nzW
+      :status: draft
+
+      It's common to use macros to avoid writing repetitive code, such as trait
+       implementations. It's possible to use derive macros or declarative macros
+       to do so.
```
jmqd, line 25:
```suggestion
      implementations. It's possible to use derive macros or declarative macros
      to do so.
```

src/coding-guidelines/macros.rst
```diff
@@ -5,3 +5,142 @@
...
+      If needing to specialize logic within the macro based on a particular
+      expression's value, it may be better to not use a declarative macro.
+
+   .. bad_example::
+      :id: bad_ex_5vK0CCmePkef
+      :status: draft
+
+      We have two macro match rules at the same level of nesting. Since the
+      matching is done macro rule by macro rule and this process is stopped as soon
+      as a macro matcher is matched we will not reach the specialized case for EmergencyValve.
```
jmqd, line 40:
```suggestion
      We have two macro match rules at the same level of nesting. Since macro
      matching is done sequentially through the matchers and stops at the first 
      match, the specialized case for EmergencyValve is unreachable.
```

Just a suggested wording change -- feel free to take it if you like it better.

src/coding-guidelines/macros.rst
```diff
@@ -5,3 +5,142 @@
...
+
+.. guideline:: Avoid specialized, fixed patterns within declarative macros
+   :id: gui_FSpI084vbwmJ
+   :status: draft
+   :fls: fls_w44hav7mw3ao
+   :tags: reduce-human-error
+   :category: macros
+   :recommendation: encouraged
+
+   Description of the guideline goes here.
```
jmqd, line 17:
```suggestion
    Matchers within macro rules are evaluated sequentially and short-circuit on
    the first match. If a specialized fixed matcher follows a broader matcher,
    it may be unreachable. This can lead to subtle and surprising bugs. It is
    encouraged to avoid the use of specialized, fixed matchers.
```

jmqd (commented):
Some in-line suggestions; I leave it up to your judgment on whether you agree / like the suggestions.

I want to think a bit on the code example. I have a feeling we could find a better example, but nothing comes to mind off the bat. Compelling code examples are indeed tricky out of context.

PLeVasseur:
@scrabsha -- I've incorporated your feedback in commit [0fda](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/pull/16/commits/0fda0d18bc1f381997f0b893c66150f01c592f1e). Thanks for calling this out.

PLeVasseur:
I think I've addressed all open feedback to the PR. Looking for final reviews before merging!

src/coding-guidelines/macros.rst
```diff
@@ -5,3 +5,151 @@
...
+    the first match. If a specialized fixed matcher follows a broader matcher,
+    it may be unreachable. This can lead to subtle and surprising bugs. It is
+    encouraged to avoid the use of specialized, fixed matchers.
+
+   .. rationale::
+      :id: rat_zqr9uEqP6nzW
+      :status: draft
+
+      It's common to use macros to avoid writing repetitive code, such as trait
+      implementations. It's possible to use derive macros or declarative macros
```
Veykril, line 28:
```suggestion
      implementations. It's possible to use procedural macros or declarative macros
```
You can very well use attributes and functional-proc macros for this as well

src/coding-guidelines/macros.rst
```diff
@@ -5,3 +5,151 @@
...
+      to do so.
+
+      In a declarative macro the ordering of the patterns will be the order that
+      they are matched against which can lead to unexpected behavior in the case
+      where we have unique behavior intended for a particular expression.
```
Veykril, line 33: > In a declarative macro the ordering of the patterns will be the order that they are matched against

This doesn't really say anything. What is the order of the patterns? As in, what do they express. I assume you are intending to describe the selection process of which rule is selected by the compiler? I'd rewrite this paragraph (not quite sure how) as the current writing is rather confusing to me (and doesn't really say much)

src/coding-guidelines/macros.rst
```diff
@@ -5,3 +5,151 @@
...
+
+      If needing to specialize logic within the macro based on a particular
+      expression's value, it is better to not use a declarative macro.
```
Veykril, line 36:
```suggestion
      If needing to specialize logic within the macro based on a particular
      expression's value, it is better to not use a declarative macro with multiple rules.
```

This change implies using distinct macros for the specialized job is fine (which imo it is)

src/coding-guidelines/macros.rst
```diff
@@ -5,3 +5,152 @@
...
+    Matchers within macro rules are evaluated sequentially and short-circuit on
+    the first match. If a specialized fixed matcher follows a broader matcher,
+    it may be unreachable. This can lead to subtle and surprising bugs. It is
+    encouraged to avoid the use of specialized, fixed matchers.
+
+   .. rationale::
+      :id: rat_zqr9uEqP6nzW
+      :status: draft
+
+      It's common to use macros to avoid writing repetitive code, such as trait
```
PLeVasseur, line 28: It would be good to explain that on refactor the ordering may be changed from:
* most specialized
* most general
to the reverse which would cause unexpected quirks later.

PLeVasseur: FYI @PLeVasseur (me) you still need to do this!

src/coding-guidelines/macros.rst
PLeVasseur: I agree 👍

src/coding-guidelines/macros.rst
PLeVasseur: @Veykril -- you're right that this is the attempt made here to describe:
> I assume you are intending to describe the selection process of which rule is selected by the compiler? 

Could I have your help with a proposed wording here?

src/coding-guidelines/macros.rst
PLeVasseur: Handled this now!

src/coding-guidelines/macros.rst
Veykril:
```

The choice of which transcriber is being used in a declarative macro depends on their declaration order within the macro. This can lead to unexpected behavior changes in invocations of declarative macros if a new transciber is inserted before another due to invocations suddenly matching a different transcriber.
```

PLeVasseur: TIL that part is called a transcriber!

src/coding-guidelines/macros.rst
```diff
@@ -28,7 +28,7 @@ Macros
       :id: non_compl_ex_Gb4zimei8cNI
       :status: draft
 
-      Explanation of code example.
+      (example of a simple expansion using a proc-macro)
```
PLeVasseur, line 31:
```suggestion
      Explanation of code example.
```

PLeVasseur:
@Veykril -- I updated with your suggested replacement using transcriber instead. Could you take a look?

felix91gr:
@pellico 

> Declarative macros should be short and easily understandable. --> **Create some complexity index, line of code, number of arguments etc...**

My 2 cents for the highlighted line: such an analysis doesn't exist. At least, not for normal code. Long story, some of it below.

I can have a chat with you to tell you about my long quest to make an implementation of SonarSource's Cognitive Complexity lint for Clippy, which then I backed from when I learned enough about (semantic) static analysis. SonarSource's lint is pure marketing and vibes, and it isn't even for macros; it's for common, procedural code.

Yet the fact that SonarSource's algorithm is fake doesn't mean that the idea of a Cognitive Complexity analysis of a piece of code is impossible. I went to bother an expert on the topic, and he was gracious enough to explain to me what the state of the art was. That was 6 years ago. He said that we were between 10 and 20 years away from having the mathematical understanding to claim to be able to make such an analysis with a straight face. And his arguments convinced me. Honestly, I now know that I underestimated what the question "Is this piece of code hard to understand?" meant.

It's much more nuanced than I thought, and answering it accurately often involves being able to ask questions that aren't even present in the code itself, like "What does the number `x` mean in this context? Is it a physical quantity? Is it perhaps a character? A ratio?". Those questions often require being able to understand documentation or to have a conversation with the programmer. But I digress.

For "normal" code, ie, procedural code, such a measurement doesn't exist yet. We don't have the mathematical machinery to perform it. That's something we won't be able to escape.

But for declarative macros... perhaps it is possible. I'm not sure. Declarative macros have an overarching structure that's very different from normal code. Perhaps it's different enough that we know how to approximate how hard they are to understand and / or maintain.

Hope that helps :')

PR #22, opened 2025-03-26
Use a local venv to allow self-modifying package installs

adfernandes:
Several of the python packages "self-modify" their own install(s), and when poetry uses a user-global-but-possibly-shared installation, that's not allowed.

The solution is to use a local `.venv` so local packages can go crazy with their installs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 __pycache__/
 build/
+.venv
--- a/poetry.toml
+++ b/poetry.toml
@@ -1,0 +1,3 @@
+[virtualenvs]
+create = true
+in-project = true


adfernandes:
@PLeVasseur - One more comment - this PR really **should** include the `poetry.lock` file generated after running `poetry update`, but I didn't want to be too intrusive, so I didn't. I can add it if you want.

Or... is there any particular reason _not_ to include the lock file? _(Adding it is the "usual standard" thing to do...)_

PLeVasseur:
Ah... yeah, because we're using `uv` not `poetry`. Can you give `uv` a spin to see if it builds?

adfernandes:
Uh oh. Is _that_ what all the cool kids are using these days? 🤣 

OMG I thought `uv` was [libuv](https://libuv.org/) for asynchronous i/o - I used it a lot years ago.

I think I'm old.

Sigh

Yes I'll give it a shot

adfernandes:
_(Grumble whine)_ worked on the first try

Sorry for the noise

PR #23, opened 2025-03-28
Normalize terminology to fit other standards

PLeVasseur:
Based on some suggestions in #19, I've tried to align to other existing standards' categories for various things

## Added

* Decidability
* Scope
* Category

## Also

* Renamed `fls.lock` to `spec.lock` to make this update a little easier later when we migrate from FLS to the upstreamed version.
* Added a flag when calling `./make.py` to allow us to update the `spec.lock` after resolving the breaking checksums

PR #24, opened 2025-03-29
Add about overflow-checks flag in release mode

iglesias:
Hi, I thought this can add a bit of information. [Flag documentation](https://github.com/rust-lang/rust/blob/master/src/doc/rustc/src/codegen-options/index.md#overflow-checks).

### DIFF
--- a/src/coding-guidelines/types-and-traits.rst
+++ b/src/coding-guidelines/types-and-traits.rst
@@ -16,8 +16,8 @@ Types and Traits
    :scope: module
    :tags: numerics
 
-   Code must not rely on Rust's implicit integer wrapping behavior that occurs in release builds. 
-   Instead, explicitly handle potential overflows using the standard library's checked, 
+   Code must not rely on Rust's implicit integer wrapping behavior that may occur in release
+   builds. Instead, explicitly handle potential overflows using the standard library's checked, 
    saturating, or wrapping operations.
 
    .. rationale::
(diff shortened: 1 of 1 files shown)


src/coding-guidelines/types-and-traits.rst
```diff
@@ -14,17 +14,18 @@ Types and Traits
...
    saturating, or wrapping operations.
 
    .. rationale:: 
       :id: rat_kYiIiW8R2qD1
       :status: draft
 
       In debug builds, Rust performs runtime checks for integer overflow and will panic if detected.
-      However, in release builds (with optimizations enabled), integer operations silently wrap
-      around on overflow, creating potential for silent failures and security vulnerabilities.
+      However, in release builds (with optimizations enabled), unless the flag overflow-checks is
```
PLeVasseur, line 26: While that will avoid overflow, I think that 'panic'ing may still be undesirable. I'd say if you added another sentence instead which described that silent wraps can be avoided with the `overflow-checks` flag, but we still `panic`, therefore we still advise being explicit that may work better. What do you think?

PLeVasseur (commented):
Thanks for adding some nuance @iglesias! Can you take a look at my comment?

PLeVasseur:
FYI @iglesias -- #23 is merged, so you can rebase this.

src/coding-guidelines/types-and-traits.rst
iglesias: Yeah, I think 'panic'inc is still undesirable in the context of this guideline's recommendation. I had considered whether it should be recommended adding it to the last sentence in the previous paragraph, then took a look at how the standard's library checked operation behaves (which is returning None) and decided not to. This change is about (probably a bit pedantically) note that the silent behavior isn't 100% implied by release / optimizations enabled. I will add another sentence here trying to clarify it.

AchimKriso:
Here is some additional context from the language team that I think might be useful:
https://lang-team.rust-lang.org/frequently-requested-changes.html#numeric-overflow-checking-should-be-on-by-default-even-in-release-mode

src/coding-guidelines/types-and-traits.rst
PLeVasseur: @iglesias -- could you update with the link that @AchimKriso shared for further context as well as the flag docs that you listed above?

PLeVasseur (commented):
Hey thanks for being so patient on this one @iglesias! I think if you can add a couple of references in here then we can merge this one in.

src/coding-guidelines/types-and-traits.rst
iglesias: Sure. Links added in https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/pull/24/commits/50ad38f5c7e3fe222beb80f0a0a9129a8170d803.

iglesias:
A check has failed here, but main seems to be good. I am going to rebase and force-push the commits.

PLeVasseur (approved):
Thank you @iglesias for adding this context and the references!

Issue #19, opened 2025-03-26
Random Notions


rcseacord:
2 Take inspiration from MISRA and CERT
2.1 Good, Bad Example => Compliant, Non-Compliant Example / Antipattern

2.2 Change “recommendation” to categories.  Categories could be Advisory, Required, Mandatory, maybe Disapplied.  Not sure what the difference between Required and Mandatory is or if it's necessary.  
2.2.1 Allow mostly only => stricter, but advisory could be loosened
3 Tightness of guidelines
3.1 “guidelines” seems to imply they are not optional

3.2 Good to use certain phrases to avoid wiggle-room

4 Examples tips
4.1 Compliant vs Non-Compliant Examples
Skirting the line on the non-compliant example helps to show where the line is.
4.2 Formatting background color of examples
Compliant => Blue
Non-Compliant => Red

To avoid fast copy-pasting of non-compliant examples.
4.3 Make sure only one thing non-compliant per example
To make clearer the issue make sure the whole example complies, except for the bit that we want to highlight.
5 Numerical Prioritization of Guidelines
Helps those retrofitting the guidelines on top of existing code to know which parts of their codebase to prioritize.
5.1 Make it numerical with 3 categories, each with a rating
6 Categories of guidelines
Useful for broad categories.
6.1 Keep loose now, tighten later
We may find that a single category should be split into N different categories.
7 Numbering
Consider some numbering schema that’s a bit more human readable in addition to the stable IDs.

Useful for tool output.
8 Tool Coverage
8.1 Consider how you will show coverage of tools over guidelines

PLeVasseur:
Thanks @rcseacord for adding some thoughts! Let's see if we can squeeze some time in tomorrow's meeting to discuss it a bit and then bring discussion back here.

PR #27, opened 2025-04-05
Add style guideline chapter

### LINKED ISSUE #25: Create strawman PR for how to read / how to write
Will help foster discussion in next week's meeting.

PLeVasseur:
* use IETF RFC 2119

* add stub for Guideline Lifecycle (status)

* add section for deprecated guidelines

* add stubbed section for Compliance (deviation process)

* add mechanism for which release the guideline is relevant for

* add about release to style guideline

closes #25


src/process/style-guideline.rst
```diff
@@ -0,0 +1,486 @@
...
+*TODO(pete.levasseur): Add more tips on when this is a good choice for a guideline.*
+
+``guideline`` ``status``
+------------------------
+
+**MUST** be one of these values:
+
+* ``draft``
+* ``approved``
+* ``deprecated``
```
PLeVasseur, line 172: Suggestion from @rcseacord to:
* remove `deprecated` as likely tool vendors will not need to issue warnings for no longer in force guidelines
* add instead a `retired` category on which to make the guideline no longer in force

PLeVasseur: Creating comment here to help solicit more feedback from Tooling Subcommittee folks 🙂

src/process/style-guideline.rst
```diff
@@ -0,0 +1,486 @@
...
+to allow adoption and feedback to accrue. The Coding Guidelines Subcommittee **MUST**
+periodically review ``draft`` guidelines and either promote them to ``approved``
+or demote them to ``deprecated``.
+
+From time to time an ``approved`` guideline **MAY** be moved to ``deprecated``. There
+could be a number of reasons, such as: a guideline which was a poor fit or wrong,
+or in order to make a single guideline more granular and replace it with
+more than one guideline.
+
+For more, see :ref:`Guideline Lifecycle`.
```
PLeVasseur, line 184: todo @PLeVasseur -- flesh this in a little before next we meet to better solicit feedback.

src/process/style-guideline.rst
```diff
@@ -0,0 +1,486 @@
...
+
+These guidelines **MUST** be enforced. Any deviations **MUST** follow the rule for their
+appropriate ``category``.
+
+``deprecated``
+^^^^^^^^^^^^^^
+
+These guidelines are not in force and **MUST** not be applied.
+
+``release``
```
PLeVasseur, line 207: Adding feedback here to ruminate on:
* use release ranges?
* earliest release may need only be as early as the first safety-certified Rust compiler
* committing to a stable release plan which is not every release for now likely fine due to slower nature of compiler adoption in safety-critical
* nature of using digital tools could allow for things like
  * having a flag to `./make.py` which would compile out certain guidelines or having a toggle within the generated docs
  * but -- should we do this? experience with MISRA seems to indicate users may not know the version and thus having a larger blanket approach may be best and not version
* handling compiler bugs usually not handled in guidelines directly, but supplementary documents such as the MISRA Compliance document for that procedure for users

PLeVasseur, line 207: We discussed having folks reach out to safety-certified Rust toolchain vendors about their roadmap so we can use this for guidance:
* HighTec - @pellico
* Ferrous Systems - @AlexCeleste
* AdaCore - @AlexCeleste

src/process/style-guideline.rst
```diff
@@ -0,0 +1,486 @@
...
+
+* ``guideline``
+* ``rationale``
+* ``non_compliant_example``
+* ``compliant_example``
+
+::
+
+   .. guideline:: Avoid Implicit Integer Wrapping
+      :id: gui_xztNdXA2oFNB
```
PLeVasseur, line 33: Suggestions on IDs from @rcseacord (some overlap with part of #19)
* something human-readable is nice
* even if wanting to stick with `gui_<alphanumeric-string>` it may be nice to help with recalling where the guideline fits by doing something like `gui_typ_<alphanumeric-string>` so that users can more easily know this is in the chapter on types and traits for example
* having identifiers like this will help with web searchability (yay)

src/process/style-guideline.rst
```diff
@@ -0,0 +1,486 @@
...
+
+A unique identifier for each guideline. Guideline identifiers **MUST** begin with ``gui_``.
+
+These identifiers are considered **stable** across releases and **MUST NOT** be removed.
+See ``status`` below for more.
+
+**MUST** be generated using the ``generate-guideline-templates.py`` script to ensure
+compliance.
+
+``category``
```
PLeVasseur, line 111: todo @PLeVasseur -- update this section with guidance that a contributor need not worry about this when initially submitting and we will discuss the right `category`.

Can help reduce friction of contributions.

src/process/style-guideline.rst
AlexCeleste: n.b. the model for tool vendors will often be along the lines of unconditionally supporting _every_ warning, with configuration left flexible to suit the user's prokect-level recategorization plan; so a category for rules that are no longer wanted by default is useful because for project-specific reasons users may want to re-enable those rules.

As a rule proceeds through its lifecycle, a tool vendor will not usually mirror that lifecycle in the corresponding warning: once any user is using a warning to enforce a rule that is Approved in their version of the document, that warning will generally be retained permanently because users do not change standard or language version rapidly, if at all, and usually find themselves bound to both a specific language revision and a specific Guidelines revision, independently of whether they can upgrade their tooling.

src/process/style-guideline.rst
```diff
@@ -0,0 +1,486 @@
...
+``approved``
+^^^^^^^^^^^^
+
+These guidelines **MUST** be enforced. Any deviations **MUST** follow the rule for their
+appropriate ``category``.
+
+``deprecated``
+^^^^^^^^^^^^^^
+
+These guidelines are not in force and **MUST** not be applied.
```
AlexCeleste, line 205: in practice this shouldn't be a *MUST*, users will continue to use deprecated guidelines for a variety of reasons. I would suggest mirroring the language for Disapplied status here and having no requirement rather than a negative requirement (i.e. all Deprecated are implicitly Disapplied regardless of recorded category)

src/process/style-guideline.rst
```diff
@@ -0,0 +1,486 @@
...
+
+An organization or project **MAY** choose to recategorize any ``advisory`` guideline as ``mandatory``
+or ``required``, or as ``disapplied``.
+
+*TODO(pete.levasseur): Add more tips on when this is a good choice for a guideline.*
+
+``disapplied``
+^^^^^^^^^^^^^^
+
+These are guidelines for which compliance **SHOULD NOT** be required. No enforcement is expected, and any
```
AlexCeleste, line 154: I think **SHOULD NOT** is too strong here, this is a non-requirement rather than a negative requirement. In the original this was just "is not required". (the **MAY** below OTOH is exactly correct)

src/process/style-guideline.rst
```diff
@@ -0,0 +1,486 @@
...
+^^^^^^^^^^
+
+A guideline which cannot be checked at the module or crate level and requires the entire
+source text **MUST** be classified as ``system``.
+
+
+``tags``
+--------
+
+The ``tags`` are largely descriptive, not proscriptive means of finding commonality between
```
AlexCeleste, line 282: typo "prescriptive" (assuming tags aren't negative :P)

src/process/style-guideline.rst
```diff
@@ -0,0 +1,486 @@
...
+a complicated concept to be conveyed using a short Title.
+
+Exception
+^^^^^^^^^
+
+Guideline Content **MAY** contain a section titled *Exception* followed by text that that describes
+situations in which the guideline does not apply. The use of exceptions permits the description of
+some guidelines to be simplified. It is important to note that an exception is a situation in which
+a guideline does not apply. Code that complies with a guideline by virtue of an exception does not
+require a deviation.
```
vapdrs, line 313: I think a distinction should be made between "does not apply" and "is not enforced" or similar wordings. The guideline on implicit integer wrapping does not apply to this example,

```rust
let sum: f32 = f + 10.0;
```

because the guideline does not apply to floating point arithmetic. No exception could be written for floating point because that is outside the guideline's scope already. For the sake of example let's say we wanted to add an exception to the implicit integer wrapping guidline for adding the integer literal `1` unchecked,

```rust
let sum = i + 1;
```

The rule **does** apply here as this is unchecked integer arithmetic prone to wraparound, but there is an explicitly provided exception which makes this specific situation OK (because we said so with some good reason). I think an exception should be categorically different than us clarifying in the description of the guideline, "This guideline only applies to integer arithmetic", which sets the scope of the rule's applicability.

```suggestion
Guideline Content **MAY** contain a section titled *Exception* followed by text that that describes specific situations where the guideline is not enforced, but would otherwise apply. The use of
exceptions permits the description of some guidelines to be simplified. It is important to note that
an exception is a situation in which compliance to the guideline is not expected. Code that complies
with a guideline by virtue of an exception does not require a deviation.
```

src/process/style-guideline.rst
```diff
@@ -0,0 +1,486 @@
...
+These identifiers are considered **stable** across releases and **MUST NOT** be removed.
+See ``status`` below for more.
+
+**MUST** be generated using the ``generate-guideline-templates.py`` script to ensure
+compliance.
+
+``rationale`` ``status``
+------------------------
+
+The ``status`` option of a ``rationale`` **MUST** match the ``status`` of its parent ``guideline``.
```
vapdrs, line 346: I'm wondering why the status of the `rationale` is separate if it must match the parent? Wouldn't it be better to instead have a `parent_id` here? The same comment applies to the following example sections

src/process/style-guideline.rst
```diff
@@ -0,0 +1,486 @@
...
+
+Rationale Content
+-----------------
+
+TODO(pete.levasseur)
```
vapdrs, line 351: Here's suggestion for kicking off this section.

```suggestion
The rationale **SHOULD** aim to explain to a reader why the guideline is useful in the context of safety critical Rust. The target audience is someone who familiar with Rust, and assumed to have read the contents of the _ The Rust Programming Language_ book, but is otherwise unaware of, or needs to be reminded about specific problematic behaviors that the guideline is attempting to cover. The rationale **MAY** include citations to specific, trusted resources that also explain the guideline, but **MUST** be able to be read coherently without accessing those resources.
```

AlexCeleste:
Conveniently, the MISRA C Working Group had a meeting today so we were able to discuss some of this PR and generate some additional comments for the review.

First and most importantly, the MISRA C WG doesn't see any IP-related risk with the currently proposed wording. What little has been copied so far clearly either falls under Fair Use or are simple dictionary terms over which MISRA does not have ownership (e.g. "Amplification" sections - not a common word - which is neat for use as a header - but we certainly didn't invent it).

Secondly, the point was made that _if_ the goal of this document is to stay in sync with the same MRAD system and Compliance etc., copying the definitions might create a maintenance burden if we change them in the C document, Obviously the C document is paywalled so it can't be used as a definition reference, but Compliance 2020 does have (slightly differently worded) definitions that can be used, and since it's free, it can be used for definitions easily. (Compliance 2020 does refer to the C document for some terms, which is unfortunate, but this will aim to be fixed in the 2025 update) So the suggestion of the group is to "as defined in Compliance 2020" for the definitions of the terms themselves, and then add longer additional descriptions in the section for guidance for people writing new Rust rules oriented purely towards helping people pick an appropriate category. This separates the concerns a little while having a single normative definition for the actual compliance case.

We like that you're doing that for the reference to the description of a deviation process!

Thirdly, someone made an interesting point that if a GRP is used to alter the category of any given Guideline, the RFC-2119 words used in the headline and Amplification might not be as-expected. We don't actually think this is a _problem_, but it might be unexpected at the user end. We don't recommend changing the wording of headlines dynamically (!), but rather having some user-facing note in the intro that RFC-2119 applies "except when a rule has been recategorized, in which case ... _etc_".
Although MISRA C doesn't do it right now, our own stylistic preference is to use **MUST** for Mandatory rules and **SHALL** for Required rules, not because the terms _or_ the enforcement of the rules are different, because they aren't; but just to emphasize visually where a rule originally came from (IMO several sentences in the PR might read better using **SHALL** but that's a purely subjective choice).

We will work to make sure that the next edition of Compliance is accessible and suitable for referencing in this way - it's already supposed to be free and language-agnostic, but we will keep Rust in mind when reviewing any updates, and ensure that minor inconsistencies like Compliance relying on any definitions from MISRA C, are completely removed. If the SCRC has any feedback about Compliance y'all want us to take into account, that would also be very welcome and will be incorporated where possible.

src/process/style-guideline.rst
PLeVasseur: Sounds good, I'll take up @rcseacord's suggestion then and incorporate it.

src/process/style-guideline.rst
PLeVasseur: Would this be covered if we had some defined things under tags like this? Such that we could have a tag underneath each guideline which denoted where it fit?

src/process/style-guideline.rst
PLeVasseur: Good idea!

src/process/style-guideline.rst
PLeVasseur: Thanks!

src/process/style-guideline.rst
PLeVasseur: Got it, thanks for providing the background.

src/process/style-guideline.rst
PLeVasseur: Thanks, will revise.

src/process/style-guideline.rst
PLeVasseur: This is a limitation of my knowledge. As far as I could tell in order to have a Need in Sphinx-Needs it's required to at a minimum have:
* an `id`
* a `status`

If we can figure out another way to do this, I'm fine with not having a `status`.

src/process/style-guideline.rst
```diff
@@ -0,0 +1,486 @@
...
+* ``required``
+* ``advisory``
+* ``disapplied``
+
+``mandatory``
+^^^^^^^^^^^^^
+
+Code claimed to be in compliance with this document **MUST** follow every guideline marked as ``mandatory``.
+
+*TODO(pete.levasseur): Add more tips on when this is a good choice for a guideline.*
```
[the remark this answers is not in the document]
PLeVasseur: Okay -- let's bring these in to discuss in the meeting next week. Thanks for the thought starters.

src/process/style-guideline.rst
```diff
@@ -0,0 +1,486 @@
...
+^^^^^^^^^^^^^^
+
+These are guidelines for which compliance **SHOULD NOT** be required. No enforcement is expected, and any
+non-compliance **MAY** be disregarded.
+
+*Note*: Where a guideline does not apply to the chosen release of the Rust compiler, it **MUST** be treated
+as ``disapplied`` for the purposes of coding guideline :ref:`Compliance`.
+
+An organization or project **MAY** choose to recategorize any ``disapplied`` guideline as ``mandatory``
+or ``required``, or as ``advisory``.
```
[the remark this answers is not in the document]
PLeVasseur: Yeah, definitely agree. Tried my best to align with MISRA C 2025's definitions. Could you help me understand if there's some action you're suggesting here?

src/process/style-guideline.rst
PLeVasseur: Can you help me understand how this changes the meaning? I've tried to align closely with MISRA C 2025's definitions.

src/process/style-guideline.rst
PLeVasseur: Done in commit [53c9973c643168f308300cffd5daf803f2d7f13b](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/pull/27/commits/53c9973c643168f308300cffd5daf803f2d7f13b)

src/process/style-guideline.rst
PLeVasseur: Done in commit [6d25a17a10dfb06ac126cf21feb9f1f34fcadcf0](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/pull/27/commits/6d25a17a10dfb06ac126cf21feb9f1f34fcadcf0)

src/process/style-guideline.rst
PLeVasseur: I think this is addressed by [1a447262fbfc9dd573abfc4e7112869269ab863e](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/pull/27/commits/1a447262fbfc9dd573abfc4e7112869269ab863e), feel free to let me know.

src/process/style-guideline.rst
PLeVasseur: I think this is addressed by [1a447262fbfc9dd573abfc4e7112869269ab863e](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/pull/27/commits/1a447262fbfc9dd573abfc4e7112869269ab863e), feel free to let me know.

src/process/style-guideline.rst
PLeVasseur: Done in [8439b0fcc162838533defc51f9c5b09e794b93a2](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/pull/27/commits/8439b0fcc162838533defc51f9c5b09e794b93a2)

PLeVasseur:
By the way this build is failing due to the `spec.lock` being out of date with the released FLS now. Addressed in #30, which should merge before this.

src/process/style-guideline.rst
vapdrs: Indeed this would cover the contents of my suggestion. However, if we are trying to find a human readable identifier for the guidelines typically I would not amalgamate the tags of something to refer to it. "You're looking for a guideline on your left near the 'integral types' and 'Undefined behavior', but if you see the 'traits' you've gone too far" :)

src/process/style-guideline.rst
vapdrs: In short I see the normative description (amplification) setting the "scope" of what does apply and what does not apply, full stop. Then the exception list would contain very specific instances of where the guideline applies, but the group agrees that it should not be enforced for some reason.

What I'm trying to prevent is the overuse of exceptions. As an exaggerated example you could instead write this guideline as "all addition shall not be used", and then provide exceptions for "float point arithmetic", "std::opts overloads", "known constant expressions", "adding zero", "operations that be statically determined to not wrap", etc.

src/retired-guidelines/index.rst
```diff
@@ -0,0 +1,30 @@
...
+   retired-attributes
+   retired-entities-and-resolution
+   retired-ownership-and-destruction
+   retired-exceptions-and-errors
+   retired-concurrency
+   retired-program-structure-and-compilation
+   retired-unsafety
+   retired-macros
+   retired-ffi
+   retired-inline-assembly
```
plaindocs, line 30: Is it useful to have all of these from the get go? Or should we just add as needed?

AlexCeleste (approved):

src/process/style-guideline.rst
```diff
@@ -0,0 +1,486 @@
...
+-------------------
+
+The Title **MUST** provide a description of the guideline.
+
+``guideline`` ``id``
+--------------------
+
+A unique identifier for each guideline. Guideline identifiers **MUST** begin with ``gui_``.
+
+These identifiers are considered **stable** across releases and **MUST NOT** be removed.
```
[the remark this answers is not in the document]
PLeVasseur: Both I'd think.

src/retired-guidelines/index.rst
PLeVasseur: I could go either way. I'm thinking to keep them for the structure to be clear.

PR #37, opened 2025-04-17
[Sphinx Extension] Include PDF build from Sphinx LaTeX

x0rw:
This is experimental at the moment, 
to generate PDFs from .rst files through Sphinx there are two approaches 
- First `rst2pdf`, it's more convinient but your control is very limited and often require direct modification to .rst files
- Second approach is the beloved LaTex, fine grained control.
the issue with customising the pdf through sphinx is the lack of documentation, learning curve, even the documentation itself is quite extensive, they really abused the word 'documentation',


x0rw:
Alright, this needs more work in the future.

senseix21:
@senseix21 is mine..

PR #48, opened 2025-04-23
[bugfix] add sphinx-autobuild to dependencies

AchimKriso:
This PR fixes the crash when running `make.py --serve` due to the missing `sphinx-autobuild` dependency.


PLeVasseur (approved):
Thanks for fixing this @AchimKriso!

PR #51, opened 2025-04-28
Add pointer type cast expression guideline: Avoid underscore casts

Veykril:


src/coding-guidelines/expressions.rst
```diff
@@ -5,3 +5,87 @@
 
 Expressions
 ===========
+
+
+as _ pointer casts. It is possible to use as _ (as well as using ptr::cast without supplying an
```
PLeVasseur, line 10: Is this paragraph here intended to allow for a grouping of guidelines around this topic?

If so -- okay, we should probably talk a bit more about organization to make sure we're doing it the same.

If not -- this should probably go under the `guideline` or the `rationale`.

src/coding-guidelines/expressions.rst
```diff
@@ -5,3 +5,87 @@
...
+explicit type argument) to cast between different raw pointer types. This should be considered an
+anti-pattern as these casts will happily accept surrounding types changing even if the cast itself
+becomes invalid in a sense. (transmute without supplying the two type arguments suffers a similar
+problem)
+
+
+::
+
+   .. guideline:: Avoid as underscore pointer casts
```
PLeVasseur, line 19: Somehow the formatting isn't rendering correctly, e.g. [here](https://safety-critical-rust-coding-guidelines-rf--51.org.readthedocs.build/en/51/coding-guidelines/expressions.html).

Could you double check?

PLeVasseur: Also the fact that the build passes despite the rendering failing make it seem to me we're not being strict enough in checking. Could you open an issue for this and take a look into this @plaindocs?

src/coding-guidelines/expressions.rst
```diff
@@ -5,3 +5,87 @@
...
+      :id: gui_HDnAZ7EZ4z6G
+      :category: required
+      :status: draft
+      :release: <TODO>
+      :fls: fls_1qhsun1vyarz
+      :decidability: decidable
+      :scope: module
+      :tags: expression, pointer
+
+      Code must not rely on Rust's type inference when doing explicit pointer casts via `var as Type` as or `core::mem::transmute`.
```
PLeVasseur, line 29: Small nit -- typo?
```suggestion
      Code must not rely on Rust's type inference when doing explicit pointer casts via `var as Type` or `core::mem::transmute`.
```

src/coding-guidelines/expressions.rst
```diff
@@ -5,3 +5,87 @@
...
+         type changes resulting in semantically invalid pointer casts.
+
+         Raw pointers have a variety of invariants to manually keep track of. Specifying the
+         concrete types in these scenarios allows the compiler to catch some of these potential
+         issues for the user.
+
+      .. non_compliant_example::
+         :id: non_compl_ex_V37Pl103aUW4
+         :status: draft
+
```
PLeVasseur, line 48: Please provide some prose for the reader to follow along with. Need not be long, just long (and short!) enough.

src/coding-guidelines/expressions.rst
```diff
@@ -5,3 +5,87 @@
...
+               let extended = extended as *const _;
+               with_base(unsafe { &*(extended as *const _) })
+            }
+
+            fn with_base(_: &Base) { ... }
+
+      .. compliant_example::
+         :id: compl_ex_W08ckDrkOhkt
+         :status: draft
+
```
PLeVasseur, line 72: Please provide some prose for the reader to follow along with. Need not be long, just long (and short!) enough.

PLeVasseur (commented):
Hey @Veykril -- thanks for filling a guideline in; recall you mentioning this as a tricky case.

Could you take a look at the comments I left?

src/coding-guidelines/expressions.rst
```diff
@@ -5,3 +5,82 @@
...
+   :tags: readability, reduce-human-error
+
+   Code must not rely on Rust's type inference when doing explicit pointer casts via `var as Type` or `core::mem::transmute`.
+   Instead, explicitly specify the complete target type in the `as` expression or `core::mem::transmute` call expression.
+
+   .. rationale::
+      :id: rat_h8LdJQ1MNKu9
+      :status: draft
+
+      `var as Type` casts and `core::mem::transmute`s between raw pointer types are generally valid and unchecked by the compiler as long the target pointer type is a thin pointer.
```
x0rw, line 27: This is what caused the build issue, the ending backtick is followed by an `s`  this is usually not an issue

x0rw: " \`\`core::mem::transmute\`\`\s  " this should work(it highlights too)

x0rw (changes requested):

src/coding-guidelines/expressions.rst
Veykril: ah right, single backticks are directives, double backticks are inline code snippets

PLeVasseur (approved):
Thank you @Veykril for bringing such a nice "gotcha"!

PR #62, opened 2025-04-30
Add guideline stub for attribute macros

AchimKriso:


PLeVasseur (approved):
@AchimKriso -- thanks for taking part in this stubbing exercise!

PR #69, opened 2025-05-07
WIP add some more proposed rules to macro guidelines

AlexCeleste:


PLeVasseur:
We chatted about this 🙂 Let's clump the Amplification together in the same group as the `guideline`. No need for another heading.

PLeVasseur:
Hey @AlexCeleste -- thanks for putting this together! Could you rebase and confirm there's no duplicates? I can then review :D

felix91gr:
Hey @AlexCeleste ^.^

What should we do about this one? Since it's been about 3 months and stuff

PR #73, opened 2025-05-08
Do not hide unsafe blocks within macro expansions

espenalb:


PLeVasseur (approved):
Thanks @espenalb for putting this skeleton in!

PR #74, opened 2025-05-11
guideline: a macro should not be used in place of a function

x0rw:
in markdown for easier reading and review:
(i'm unsure about the proper term for every italicized word below)

# A macro should not be used in place of a function 
   :category: _mandatory_
   :tags: reduce-human-error
   :decidability: decidable

   Functions should always be preferred over macros, except when macros provide essential functionality that functions cannot, such as _variadic interfaces_, compile-time code generation, or _syntax extensions_ via custom derive and attribute macros.

## Rationale:
Macros are powerful but they come at the cost of readability, complexity, and maintainability. They _obfuscate_ control flow and type signatures,

 **Debugging Complexity** 
- Errors point to expanded code rather than source locations, making it difficult to trace compile-time errors back to the original macro invocation.

**Optimization**   
- Macros **may** inhibit compiler optimizations that work better with functions.
- Macros act like ``#[inline(always)]`` functions, which can lead to code bloat.
- They don't benefit from the compiler's [inlining heuristics](https://doc.rust-lang.org/nightly/reference/attributes/codegen.html), missing out on selective inlining where the compiler decides when inlining is beneficial.

**Functions provide**

- Clear type signatures.
- Predictable behavior.
- Proper stack traces.
- _Consistent_ optimization opportunities.

### DIFF
--- a/src/coding-guidelines/macros.rst
+++ b/src/coding-guidelines/macros.rst
@@ -90,6 +90,71 @@ Macros
 
         // TODO
 
+.. guideline:: a macro should not be used in place of a function 
+   :id: gui_2jjWUoF1teOY
+   :category: mandatory
+   :status: draft
+   :release: todo
+   :fls: fls_xa7lp0zg1ol2
+   :decidability: decidable
+   :scope: system
+   :tags: reduce-human-error
+
+   Functions should always be preferred over macros, except when macros provide essential functionality that functions cannot, such as variadic interfaces, compile-time code generation, or syntax extensions via custom derive and attribute macros.
+    
+   |
+
+   .. rationale:: 
+      :id: rat_M9bp23ctkzQ7
+      :status: draft
+
+      Macros are powerful but they come at the cost of readability, complexity, and maintainability. They obfuscate control flow and type signatures,
+
+      **Debugging Complexity** 
+
+      - Errors point to expanded code rather than source locations, making it difficult to trace compile-time errors back to the original macro invocation.
+
+      **Optimization**
+      
+      - Macros may inhibit compiler optimizations that work better with functions.
+      - Macros act like ``#[inline(always)]`` functions, which can lead to code bloat.
+      - They don't benefit from the compiler's inlining heuristics, missing out on selective inlining where the compiler decides when inlining is beneficial.
+
+      **Functions provide**
+
+      - Clear type signatures.
+      - Predictable behavior.
+      - Proper stack traces.
+      - Consistent optimization opportunities.
+
+
+   .. non_compliant_example::
+      :id: non_compl_ex_TZgk2vG42t2r
+      :status: draft
+
+      Using a macro where a simple function would suffice:
+   
+      .. code-block:: rust
+   
+        macro_rules! add {
+          ($a:expr, $b:expr) => { $a + $b };
+        }
+
+   .. compliant_example::
+      :id: compl_ex_iPTgzrvO7qr3
+      :status: draft
+
+      Implementing the same functionality as a function:
+
+   
+      .. code-block:: rust
+
+        fn add(a: i32, b: i32) -> i32 { 
+          a + b 
+        }
+
+        let sum = add(2, 32);  // Clear, type-safe, and debuggable 
+
 .. guideline:: Shall not use Function-like Macros
    :id: gui_WJlWqgIxmE8P
    :category: mandatory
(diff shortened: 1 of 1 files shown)


src/coding-guidelines/macros.rst
```diff
@@ -90,6 +90,71 @@ Macros
 
         // TODO
 
+.. guideline:: a macro should not be used in place of a function
```
PLeVasseur, line 93:
```suggestion
.. guideline:: A macro should not be used in place of a function
```

src/coding-guidelines/macros.rst
```diff
@@ -90,6 +90,71 @@ Macros
...
+   :status: draft
+   :release: todo
+   :fls: fls_xa7lp0zg1ol2
+   :decidability: decidable
+   :scope: system
+   :tags: reduce-human-error
+
+   Functions should always be preferred over macros, except when macros provide essential functionality that functions cannot, such as variadic interfaces, compile-time code generation, or syntax extensions via custom derive and attribute macros.
+    
+   |
```
PLeVasseur, line 105: I'm not familiar enough. What purpose does the `|` serve?

src/coding-guidelines/macros.rst
```diff
@@ -90,6 +90,71 @@ Macros
...
+
+   .. rationale:: 
+      :id: rat_M9bp23ctkzQ7
+      :status: draft
+
+      Macros are powerful but they come at the cost of readability, complexity, and maintainability. They obfuscate control flow and type signatures,
```
PLeVasseur, line 111:
```suggestion
      Macros are powerful but they come at the cost of readability, complexity, and maintainability. They obfuscate control flow and type signatures.
```

PLeVasseur (commented):
Thank @x0rw for adding this. Could you take a look at some of the comments?

src/coding-guidelines/macros.rst
x0rw: It's usually used for escaping characters, but here i'm using it as a 'line block', to add an extra line instead of using raw html, without an extra line the lines look crammed.

For reference [check this out](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#line-blocks)

PLeVasseur: Ah gotcha. Thanks for the reference.

src/coding-guidelines/macros.rst
```diff
@@ -90,6 +90,71 @@ Macros
...
+      - Predictable behavior.
+      - Proper stack traces.
+      - Consistent optimization opportunities.
+
+
+   .. non_compliant_example::
+      :id: non_compl_ex_TZgk2vG42t2r
+      :status: draft
+
+      Using a macro where a simple function would suffice:
```
PLeVasseur, line 135: I think it could be useful here to show where usage of the `add!()` macro could lead to some of the disadvantages you listed above.

If it's not possible to tie this specific example back to those disadvantages, it may be needed to rework the non-compliant example.

src/coding-guidelines/macros.rst
```diff
@@ -90,6 +90,71 @@ Macros
...
+        macro_rules! add {
+          ($a:expr, $b:expr) => { $a + $b };
+        }
+
+   .. compliant_example::
+      :id: compl_ex_iPTgzrvO7qr3
+      :status: draft
+
+      Implementing the same functionality as a function:
+
```
PLeVasseur, line 148: Extra line

```suggestion
```

src/coding-guidelines/macros.rst
```diff
@@ -90,6 +90,71 @@ Macros
...
+   
+        macro_rules! add {
+          ($a:expr, $b:expr) => { $a + $b };
+        }
+
+   .. compliant_example::
+      :id: compl_ex_iPTgzrvO7qr3
+      :status: draft
+
+      Implementing the same functionality as a function:
```
PLeVasseur, line 147: It would be good to highlight how using a function here avoids the disadvantages you've listed above.

PLeVasseur (commented):
Thanks @x0rw! Please take a look at the comments I left

src/coding-guidelines/macros.rst
x0rw: Indeed, i picked another example, it's my favorite and it shows the disadvantages of using macros instead of functions, please check it out.

PLeVasseur: Great example!

PR #76, opened 2025-05-14
New guideline to refer to names with global paths in a macro

vapdrs:
As discussed in the 2025-05-07 meeting, a draft of this guideline was asked to be contributed. I'd be interested in any feedback, specifically for the release section, whose specification still seems in flux, and the rationale which I generated but am not sure if it captures the original idea contributor's purpose.

### DIFF
--- a/src/coding-guidelines/macros.rst
+++ b/src/coding-guidelines/macros.rst
@@ -427,9 +427,9 @@ Macros
             // Compliant implementation
         }
 
-.. guideline:: Names in a macro definition shall use a global path
+.. guideline:: Names in a macro definition shall use a fully qualified path
    :id: gui_SJMrWDYZ0dN4
-   :category: advisory
+   :category: required
    :status: draft
    :release: 1.85.0;1.85.1
    :fls: fls_7kb6ltajgiou
@@ -437,15 +437,15 @@ Macros
    :scope: module
    :tags: reduce-human-error
 
-   Each name inside of the definition of a macro shall use a global path, which absolutely refers to an
-   entity.
+   Each name inside of the definition of a macro shall either use a global path or path prefixed with $crate.
 
    .. rationale::
       :id: rat_VRNXaxmW1l2s
       :status: draft
 
       Using a path that refers to an entity relatively inside of a macro subjects it to path resolution
-      results which may change depending on where the macro is used. This could lead to developer confusion
+      results which may change depending on where the macro is used. The intended path to refer to an entity
+      can be shadowed when using a macro leading to unexpected behaviors. This could lead to developer confusion
       about why a macro behaves differently in diffenent locations, or confusion about where entity in a macro
       will resolve to.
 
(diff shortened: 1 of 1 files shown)


src/coding-guidelines/macros.rst
```diff
@@ -342,4 +342,69 @@ Macros
 
         fn example_function() {
             // Compliant implementation
-        }
\ No newline at end of file
+        }
+
+.. guideline:: Names in a macro definition shall use a global path
+   :id: gui_SJMrWDYZ0dN4
+   :category: advisory
```
x0rw, line 349: Hey :wave:.
I think it should be `mandatory` because even a local `use` can easily shadow the intended path and cause unexpected behaviours,
Also not following this guideline will make your nacro non-deterministic and context-dependent.

vapdrs: Thanks for the review. I'm willing to move this to `required`, especially if the `$crate` path prefix is also allowed for. 

Moving it further to `mandatory` feels a bit strong, since I feel like there are exotic use cases for shadowing the paths inside of the macro. Right now I can't think of any patterns that use that though, so perhaps there aren't. I've updated category in 6698d44, feel free to let me know if you still think this should be mandatory.

x0rw: Yes 'mandatory' feels quite strong, maybe this should be discussed in the next meeting.

src/coding-guidelines/macros.rst
```diff
@@ -342,4 +342,69 @@ Macros
...
+      results which may change depending on where the macro is used. The intended path to refer to an entity
+      can be shadowed when using a macro leading to unexpected behaviors. This could lead to developer confusion
+      about why a macro behaves differently in diffenent locations, or confusion about where entity in a macro
+      will resolve to.
+
+   .. non_compliant_example::
+      :id: non_compl_ex_m2XR1ihTbCQS
+      :status: draft
+
+      The following is a macro which shows referring to a vector entity using a non-global path.
```
PLeVasseur, line 373: It might be nice to explain how depending on where the macro is used, if there are different `Vec` in scope, this could result in different behavior.

Maybe we don't need more code to show it, but simply explain it in a sentence or two.

src/coding-guidelines/macros.rst
```diff
@@ -342,4 +342,69 @@ Macros
...
+                    temp_vec
+                }
+            };
+        }
+
+   .. compliant_example::
+      :id: compl_ex_xyaShvxL9JAM
+      :status: draft
+
+      The following is a macro refers to Vec using a global path.
```
PLeVasseur, line 394: Similarly here we could explain that even if in our own crate or using a third party crate if there is a type `Vec` by doing this fully-qualified path we avoid accidentally referring to either of those types and ensure we refer only to the `std` version.

PLeVasseur (commented):
Hey @vapdrs -- thanks for contributing this guideline! Please take a look at the comments I left.

PLeVasseur:
@vapdrs -- could you rebase? Then we can spin CI

PLeVasseur (approved):
Thanks for putting this together @vapdrs! Looks good to me

PR #84, opened 2025-05-21
Rework the guideline description to remove the incorrect claim that a…

AlexCeleste:
…n explicit Amplification section exists (instead, whatever paragraph immediately opens the guideline is just normative). We can continue to call it that but the subheading is not necessary.

### DIFF
--- a/src/process/style-guideline.rst
+++ b/src/process/style-guideline.rst
@@ -92,10 +92,6 @@ We will examine each part:
       :scope: module
       :tags: numerics
 
-      Code must not rely on Rust's implicit integer wrapping behavior that occurs in release builds.
-      Instead, explicitly handle potential overflows using the standard library's checked,
-      saturating, or wrapping operations.
-
 ``guideline`` Title
 -------------------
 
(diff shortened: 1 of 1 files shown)


src/process/style-guideline.rst
```diff
@@ -92,10 +92,6 @@ We will examine each part:
       :scope: module
       :tags: numerics
 
-      Code must not rely on Rust's implicit integer wrapping behavior that occurs in release builds.
```
PLeVasseur, line 95: Ah, this is removed here in order to explain it under the Amplification section?

src/process/style-guideline.rst
```diff
@@ -363,7 +379,13 @@ The ``status`` option of a ``rationale`` **MUST** match the ``status`` of its pa
 Rationale Content
 -----------------
 
-TODO(pete.levasseur)
+The content of the rationale **SHOULD** provide the relevant context for why this guideline is useful.
+The Rationale **SHOULD** make reference to any undefined behaviours or known errors associated
```
PLeVasseur, line 383: What say we? British English or American English? 😆 

```suggestion
The Rationale **SHOULD** make reference to any undefined behaviors or known errors associated
```

src/process/style-guideline.rst
```diff
@@ -308,15 +304,27 @@ what it covers.
 Content **SHOULD** aim to be as short and self-contained as possible, while still explaining
 the scope of the guideline.
 
-Content **SHOULD NOT** cover the rationale for the guideline, which is done in the ``rationale`` section.
+Guideline content consists of an Amplification and any Exceptions, which are normative,
```
PLeVasseur, line 307: Oh, is this the reason for removing the content up above in the example `guideline`?

Any text under the `guideline` should be _either_ an Amplification or an Exception?

PLeVasseur (commented):
Hey @AlexCeleste -- I left comments to make sure I understand the thrust of what you've written. Thanks for taking a stab at this!

PR #88, opened 2025-05-23
updated the base URL in fls_checks.py

Strange-Satwik:
Updated the base URL in fls_checks.py

resolves https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/issues/87


PLeVasseur (approved):
Thanks @Strange-Satwik!

PR #126, opened 2025-06-04
Fix rationale for "A macro should not be used in place of a function" 

### LINKED ISSUE #90: Fix rationale for "A macro should not be used in place of a function"
In  "A macro should not be used in place of a function" guideline's rational, there is a misleading line under the `Debugging complexity` section, 
> **Debugging Complexity: **
>- Errors point to expanded code rather than source locations, making it difficult to trace compile-time errors back to the original macro invocation.

The error points to the macro and it's invocation source as well.

## Why
Macros are expanded before reaching the HIR and every expanded token is coupled with a [Span](https://doc.rust-lang.org/beta/nightly-rustc/rustc_span/struct.Span.html), this `span` keeps track of the macro and it's invocation site as well, this span is also carried through nested macros, 
example: If a macro A expands into code that invokes another macro B, and a problematic token originates within B's expansion, the token's span allows tracing its origin. This means you can refer to B's invocation location (which is inside the code generated by A), then back to A's macro code itself, and finally to A's original invocation source in your file.
check this playground: [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=fc45a717acad42e1783ba7cd6a4df109)

>    = note: this error originates in the macro `macro_B` which comes from the expansion of the macro `macro_A`

## Change
Change ~Errors point to expanded code rather than source locations, making it difficult to trace compile-time errors back to the original macro invocation~ .
to: "Although the compiler reports both the macro expansion and its invocation site, diagnostics originating within macros can be more difficult to interpret than those from ordinary function or type definitions. Complex or deeply nested macros may obscure intent and hinder static analysis, increasing the risk of misinterpretation or overlooked errors during code review."

### Side note:
Procedural macros allow you to control "Spans", and this control comes with the cost of increased complexity for the macro author, and a greater responsibility to generate correct and hygienic code that produces helpful diagnostics.

### references:
https://doc.rust-lang.org/beta/nightly-rustc/rustc_span/index.html
https://rustc-dev-guide.rust-lang.org/diagnostics.html#span
https://doc.rust-lang.org/beta/nightly-rustc/rustc_span/hygiene/struct.SyntaxContext.html

x0rw:
Changed `Errors point to expanded code rather than source locations, making it difficult to trace compile-time errors back to the original macro invocation`.

to: `Although the compiler reports both the macro expansion and its invocation site, diagnostics originating within macros can be more difficult to interpret than those from ordinary function or type definitions. Complex or deeply nested macros may obscure intent and hinder static analysis, increasing the risk of misinterpretation or overlooked errors during code review`.

Check #90 for more informations.
Resolves #90.

### DIFF
--- a/src/coding-guidelines/macros.rst
+++ b/src/coding-guidelines/macros.rst
@@ -108,7 +108,8 @@ Macros
       :id: rat_M9bp23ctkzQ7
       :status: draft
 
-      Macros are powerful but they come at the cost of readability, complexity, and maintainability. They obfuscate control flow and type signatures.
+      Although the compiler reports both the macro expansion and its invocation site, diagnostics originating within macros can be more difficult to interpret than those from ordinary function or type definitions. Complex or deeply nested macros may obscure intent and hinder static analysis, increasing the risk of misinterpretation or overlooked errors during code review.
+     
 
       **Debugging Complexity** 
 


PLeVasseur (approved):
Thanks for the revision @x0rw!

PR #127, opened 2025-06-04
Defects vs subsets

AlexCeleste:
This adds a basic example of three kinds of guideline:

- a blanket Advisory subsetting guideline
- a more targeted decidable Required subsetting guideline
- a defect-oriented undecidable Required guideline


src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,245 @@ Expressions
...
+      :id: non_compl_ex_0ydPk7VENSrA
+      :status: draft
+
+      Any use of ``as`` or ``transmute`` to create a pointer from an arithmetic address value
+      is non-compliant:
+
+      .. code-block:: rust
+
+        fn f1 (x:u16, y:i32, z:u64, w:usize) {
+          let p1 = x as * const u32;  // not compliant
```
sei-dsvoboda, line 239: Can NULL (0) be safely converted from integer to pointer in Rust?

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,245 @@ Expressions
...
+        }
+
+   .. compliant_example::
+      :id: compl_ex_oneKuF52yzrx
+      :status: draft
+
+      There is no compliant example of this operation.
+
+
+.. guideline:: An integer shall not be converted to an invalid pointer
```
sei-dsvoboda, line 265: I take it this is a variant or subset of the previous rule? A dev should choose to follow this rule or the prev one, right?

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,245 @@ Expressions
...
+   :status: draft
+   :release: <TODO>
+   :fls: fls_9wgldua1u8yt
+   :decidability: undecidable
+   :scope: system
+   :tags: defect, undefined-behavior
+
+   An expression of numeric type shall not be converted to a pointer if the resulting pointer
+   is incorrectly aligned, does not point to an entity of the referenced type, or is an invalid representation.
+
```
sei-dsvoboda, line 277: I recognize this wording from CERT INT36-C :)   IIRC it all comes from ISO C. That aside, does the above sentence have the same meaning in Rust?  Or is Rust relying on the ISO C definitions of these terms?

src/process/style-guideline.rst
```diff
@@ -308,15 +304,27 @@ what it covers.
 Content **SHOULD** aim to be as short and self-contained as possible, while still explaining
 the scope of the guideline.
 
-Content **SHOULD NOT** cover the rationale for the guideline, which is done in the ``rationale`` section.
```
PLeVasseur, line 311: I think you mentioned about removing the IETF RFC 2119-styling to this chapter when it was intended for readers. Now that it's intended for contributors are you thinking to keep it in?

Or is this change unrelated to that to aid in understanding?

src/process/style-guideline.rst
```diff
@@ -363,7 +379,13 @@ The ``status`` option of a ``rationale`` **MUST** match the ``status`` of its pa
 Rationale Content
 -----------------
 
-TODO(pete.levasseur)
+The content of the rationale **SHOULD** provide the relevant context for why this guideline is useful.
```
PLeVasseur, line 382: Should this be **MUST** or **SHOULD**? Reading through IETF RFC 2119 I debated myself.

Same for the one below on undefined behavior.

> [**MUST**](https://datatracker.ietf.org/doc/html/rfc2119#section-1)   This word, or the terms "REQUIRED" or "SHALL", mean that the definition is an absolute requirement of the specification.

> [**SHOULD**](https://datatracker.ietf.org/doc/html/rfc2119#section-3)   This word, or the adjective "RECOMMENDED", mean that there may exist valid reasons in particular circumstances to ignore a particular item, but the full implications must be understood and carefully weighed before choosing a different course.

src/process/style-guideline.rst
```diff
@@ -363,7 +379,13 @@ The ``status`` option of a ``rationale`` **MUST** match the ``status`` of its pa
...
+The Rationale **SHOULD** make reference to any undefined behaviors or known errors associated
+with the subject of the guideline.
+The Rationale **MAY** make reference to other guidelines or to external documents cited in the
+References.
+
+The Rationale **SHOULD** be supported by code examples wherever concise examples are possible.
```
PLeVasseur, line 388: This is a nice usage of **SHOULD** 👍

src/process/style-guideline.rst
```diff
@@ -435,6 +457,9 @@ than the current guideline.
 ``compliant_example``
 =====================
 
+A compliant example **MAY** be omitted when the guideline forbids an action entirely, i.e. there
```
PLeVasseur, line 460: Should this be **SHOULD** to encourage the right action taken by a guideline writer? 🤔 
```suggestion
A compliant example **MAY** be omitted when the guideline forbids an action entirely, i.e. there
```

PLeVasseur (commented):
Thanks @AlexCeleste for the excellent additions! Could you take a look at the comments I left?

By the way -- I think I recall you saying the plan is to incorporate these examples more into the Style Guideline to help flesh it out. Was that right? In this PR or another?

I suppose then it would be quite helpful to use the examples to illustrate the difference between a defect and subset.

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,245 @@ Expressions
...
+      :id: rat_v56bjjcveLxQ
+      :status: draft
+
+      Although the conversions performed by ``as`` between numeric types are all well-defined, ``as`` coerces
+      the value to fit in the destination type, which may result in unexpected data loss if the value needs to
+      be truncated, rounded, or produce a nearest possible non-equal value.
+
+      Although some conversions are lossless, others are not symmetrical. Instead of relying on either a defined
+      lossy behaviour or risking loss of precision, the code can communicate intent by using ``Into`` or ``From``
+      and ``TryInto`` or ``TryFrom`` to signal which conversions are intended to perfectly preserve the original
```
onixie, line 117: `TryInto` and `TryFrom` cannot be used in const function yet. 
Don't know if it can be considered as an exception but at least it can be good to mention as a note in the guideline. 

And if we can recommend lints in this guideline, I believe 
- `cast_possible_truncation`
- `cast_possible_wrap`
- `cast_precision_loss`
- `cast_enum_truncation`
should be turned on for numeric conversions.

iglesias (approved):
I've checked the additions a few times over the past weeks and they look quite good to me overall.

src/conf.py
```diff
@@ -99,6 +99,10 @@
     dict(name="readability", description="Readability-related guideline"),
     dict(name="reduce-human-error", description="Reducing human error guideline"),
     dict(name="numerics", description="Numerics-related guideline"),
+    dict(name="undefined-behavior", description="Numerics-related guideline"),
+
+    dict(name="defect", description="Guideline associated with the language-subset profile"),
```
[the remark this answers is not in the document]
PLeVasseur: Created an issue for this, perhaps we tackle this later:
https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/issues/142

What do you think @AlexCeleste?

src/process/style-guideline.rst
PLeVasseur: Perhaps this was rewritten in order to read better? Curious to hear your thoughts.

PLeVasseur (commented):
Left a couple of thoughts! Thanks a bunch for sweeping through and addressing comments @AlexCeleste

PLeVasseur (changes requested):
Missed a comment. Realized we can also use the linking to the standard library for `core` and other standard library crates as well 🥳

src/conf.py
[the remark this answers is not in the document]
PLeVasseur: Let's tackle later. Marking as resolved since we have an issue now.

PLeVasseur (commented):
I think we're almost there! Last question about choice of ETF RFC 2119-styling on one section.

PLeVasseur (commented):
Also -- seems as though the double backtick syntax for Guideline title isn't working here:

image

Could you either remove the backticks or look into how to get this to still render correctly as inline code?

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,245 @@ Expressions
...
+
+           let g = e as i64;            // non-compliant - lossy despite object size
+
+           let p1:* const u32 = &b;
+           let a1 = p1 as usize;        // compliant by exception
+           let a2 = p1 as u16;          // non-compliant - may lose address range
+           let a3 = p1 as u64;          // non-compliant - use usize to indicate intent
+
+           let p2 = a1 as * const u32;  // non-compliant - prefer transmute
+           let p3 = a2 as * const u32;  // non-compliant, and probably invalid
```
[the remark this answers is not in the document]
PLeVasseur: Gotcha. Thanks for detailing the reason. 👍

src/process/style-guideline.rst
PLeVasseur: Thanks for the reasoning 👍

PLeVasseur (approved):
Thanks @AlexCeleste for these excellent examples and fleshing out the style guidelines!

Issue #75, opened 2025-05-14
MISRA vs CERT


rcseacord:
The MISRA and CERT Coding Standards have significantly different philosophies.  One of these is correct, and the other is MISRA.

MISRA may seem more popular in certain circles, but this is misleading.  MISRA has been around much longer and has been successfully tested in court as a liability defense, while the Safety Critical Rust Coding Guidelines have not. 

CERT is designed to work with existing code to improve security.  Safety and security are closely related.  Security often has more requirements, for example, that your credit card information is not leaked.  Leaking of information is not a safety concern.

The CERT requirement to work with existing code has several consequences.  CERT does not attempt to subset the language.  Instead, CERT focuses on actual defects in code that can result in security failures and vulnerabilities.  CERT does not require developers to follow unproven coding rules from the 1970s that someone had once thought were a good idea but was not backed by any proven research or data and has been since disproven.

MISRA takes a significantly different approach.  The MISRA approach is subset the language and then make you explain why it's OK for you code in the language.  They create many rules which maybe at some time seemed like a good idea but for which there is insufficient evidence that these rules actually improve your code.  Many of these rules are based on early work from Les Hatton [Hatton 1995]   The problem here is that Less Hatton was a smart guy and most of what he said made sense at the time even if some of it has since been contradicted.  Unfortunately, the MISRA Working Groups with the exception of a couple people such as Alex are not very smart people, are inflexible in their thinking, and wed to outdated dogma.

There are many theories of how MISRA is supposed to work, many of which are incorrect.  The most misleading of these theories is that MISRA defines a safe, analyzable subset of the C or C++ programming language.  MISRA understands that you might not be able to implement your system using this language subset, so they define a deviation process.  This let's you deviate from most (non-mandatory) rules, simply by doing a lot of extra, unnecessary work.  Basically you have to create an safety argument to convince an auditor that you've thought about it.  But this process will ensure a safe system.
Many of these theories are put forth in _MISRA Compliance:2020 Achieving compliance with MISRA Coding Guidelines_, published in February 2020.  This document comes with a large disclaimer:

> Disclaimer
> 
> Adherence to the requirements of this document does not in itself ensure error-free robust software or guarantee portability and re-use.
> 
> Compliance with the requirements of this document, or any other standard, does not of itself confer immunity from legal obligations.

[...]

PR #132, opened 2025-06-10
[Guideline] Add do not divide by 0

### LINKED ISSUE #131: [Coding Guideline]: Do not divide by 0
### Chapter

Expressions

### Guideline Title

Do not divide by 0

### Category

Mandatory

### Status

Draft

### Release Begin

unclear

### Release End

latest

### FLS Paragraph ID

fls_Q9dhNiICGIfr

### Decidability

Undecidable

### Scope

System

### Tags

numerics

### Amplification

This guideline applies when unsigned integer or two’s complement division is performed.

### Exception(s)

_No response_

### Rationale

Integer division by zero results in a panic, which is an abnormal program state and may terminate the process.

### Non-Compliant Example - Prose

When the division is performed, the right operand is evaluated to zero and the program panics.

### Non-Compliant Example - Code

let x = 0;
let x = 5 / x;

### Compliant Example - Prose

There is no compliant way to perform integer division by zero

### Compliant Example - Code

let x = 5 / 5;

vapdrs:
Closes #131


felix91gr:
Hi @vapdrs! I have already sent you a message in Zulip, but here seems like a better place to do so.

I come to add a couple of things to this guideline :)

1) There is a Clippy lint (actually a couple of lints, but this one is the main one) that one can enable to catch these kinds of operations: [arithmetic_side_effects](https://rust-lang.github.io/rust-clippy/master/index.html?groups=restriction#arithmetic_side_effects). This is a fairly general lint that, [if enabled](https://github.com/rust-lang/rust-clippy/?tab=readme-ov-file#allowingdenying-lints) using `warn` or `deny`, lints any arithmetic expression that either [overflows](https://doc.rust-lang.org/reference/expressions/operator-expr.html#overflow) or panics (division and modulo by 0 will do the latter).

2) The aforementioned lint should indicate to the user that there are arithmetic operations in the stdlib that can be used to guarantee well-defined behavior. In the case of division by zero, it will suggest using operations such as [`checked_div`](https://docs.rs/num/latest/num/trait.CheckedDiv.html#tymethod.checked_div), which outputs an `Option<T>` that the user must then handle properly. The `None` result indicates overflow, underflow or division by zero.

There are other such operations for division, such as [`checked_div_rem_euclid`](https://docs.rs/num/latest/num/traits/trait.CheckedEuclid.html#method.checked_div_rem_euclid) for when a remainder is desired.

3) A complement to the two resources above is the [NonZero](https://doc.rust-lang.org/std/num/struct.NonZero.html) type. One uses NonZero<T> to enclose a value that is known, by construction, to not equal zero.

This combines rather well with Option, as in Option<NonZero<T>>, since the compiler can do some memory layout optimization due to the fact that the value being enclosed by NonZero has one bit-pattern that is known to not be possible (the 000...000 pattern)

I will review the PR shortly :3

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,43 @@ Expressions
...
+   :tags: numerics
+
+   This guideline applies when unsigned integer or two’s complement division is performed. This includes the
+   evaluation of a remainder expression.
+
+   .. rationale::
+      :id: rat_h84NjY2tLSBW
+      :status: draft
+
+      Integer division by zero results in a panic, which is an abnormal program state and may terminate the process.
```
felix91gr, line 102: Indeed. It's even worse than that. Division by zero is Undefined Behavior, which is instant death for Safety Critical purposes :3

PLeVasseur (commented):
Thanks for opening this up @vapdrs! Could you check the comments I left?

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,43 @@ Expressions
...
 
          fn with_base(_: &Base) { ... }
+
+.. guideline:: Do not divide by 0
+   :id: gui_kMbiWbn8Z6g5
+   :category: Mandatory
+   :status: draft
+   :release: latest
+   :fls: fls_Q9dhNiICGIfr
+   :decidability: Undecidable
```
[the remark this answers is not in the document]
vapdrs: I think this is a good suggestion. As discussed in #75, I think a good way forward would be to leave the categorization of this guideline Mandatory and Undecidable, and also create a Required, Decidable rule which enforces "Always use checked arithmetic". Compliance with the required rule would ensure compliance with the mandatory rule. I'll create that other rule in a separate PR.

The reason I think this approach makes sense is because of

> Does that make this quite onerous to the engineer which must comply with this?

If you are starting off a new project, with new code then it should be easy to comply with a "always used checked arithmetic rule".

If instead you are inheriting a large project that now needs to become compliant with our guidelines, the safety engineer can pick their poison: Either do a large rewrite to comply with "always used checked arithmetic rule" to ensure you never divide by zero, or do a manual review and justification at every location where static analysis can't determine you aren't dividing by zero.

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,43 @@ Expressions
...
+      .. code-block:: rust
+
+         let x = 0;
+         let x = 5 / x;
+
+   .. compliant_example::
+      :id: compl_ex_Ri9pP5Ch3kbb
+      :status: draft
+
+      There is no compliant way to perform integer division by zero
```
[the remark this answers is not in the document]
vapdrs: Resolved in a976035

src/coding-guidelines/expressions.rst
adfernandes: **Belay that!**

I just checked the assembly for `thumbv8m`, and the code generator and optimizer _remove the check_!

ASM snippet, see the `udiv` instruction:

```
00000ad8 <$t>:
00000ad8 <unicorn::sort::i32::hec332188e2428c44>:
00000ad8 <_entry>:
     ad8: b580         	push	{r7, lr}
     ada: af00         	add	r7, sp, #0x0
     adc: b173         	cbz	r3, 0xafc <_entry+0x24> @ imm = #0x1c
     ade: fbb2 f2f3    	udiv	r2, r2, r3
     ae2: 2a2a         	cmp	r2, #0x2a
     ae4: d001         	beq	0xaea <_entry+0x12>     @ imm = #0x2
     ae6: 2902         	cmp	r1, #0x2
     ae8: d200         	bhs	0xaec <_entry+0x14>     @ imm = #0x0
     aea: bd80         	pop	{r7, pc}

```

adfernandes: Just for reference:
* [llvm's udiv instruction](https://llvm.org/docs/LangRef.html#udiv-instruction)
* [llvm's undefined behavior](https://llvm.org/docs/UndefinedBehavior.html)

vapdrs: Very neat, thank you for looking! Perhaps this is worth bringing up to the {opsem team](https://github.com/rust-lang/opsem-team) Félix mentioned in the meeting today.

I'm not proficient in `thumbv8m`, but if `cbz` stands for "compare branch zero" I'd be interested in what it is comparing and where it branches, to see if that is guarding the `udiv`.

adfernandes: @vapdrs, I went a little too quickly and built my object in "relocatable" mode, which buggered me up.

The zero check is **not** elided.

The complete assembly is below. It's a little convoluted because of how I had to build the minimal test case (apologies).

The `10000aac cbz` **does indeed** do the zero check and branches to `10000b02`, which loads the message and branches to the `panic_const_div_by_zero` function.

Again, this is at `opt=s` optimization.

```
10000aa8 <$t>:
10000aa8 <unicorn::sort::i32::hec332188e2428c44>:
10000aa8 <_entry>:
10000aa8: b5d0         	push	{r4, r6, r7, lr}
10000aaa: af02         	add	r7, sp, #0x8
10000aac: b34b         	cbz	r3, 0x10000b02 <_entry+0x5a> @ imm = #0x52
10000aae: fbb2 f2f3    	udiv	r2, r2, r3
10000ab2: 2a2a         	cmp	r2, #0x2a
10000ab4: d020         	beq	0x10000af8 <_entry+0x50> @ imm = #0x40
10000ab6: 2902         	cmp	r1, #0x2
10000ab8: bf38         	it	lo
10000aba: bdd0         	poplo	{r4, r6, r7, pc}
10000abc: 2915         	cmp	r1, #0x15
10000abe: d21c         	bhs	0x10000afa <_entry+0x52> @ imm = #0x38
10000ac0: eb00 0c81    	add.w	r12, r0, r1, lsl #2
10000ac4: 1d02         	adds	r2, r0, #0x4
10000ac6: f04f 0e04    	mov.w	lr, #0x4
10000aca: e952 4101    	ldrd	r4, r1, [r2, #-4]
10000ace: 42a1         	cmp	r1, r4
10000ad0: da0d         	bge	0x10000aee <_entry+0x46> @ imm = #0x1a
10000ad2: 4673         	mov	r3, lr
10000ad4: 2b04         	cmp	r3, #0x4
10000ad6: 50c4         	str	r4, [r0, r3]
10000ad8: d007         	beq	0x10000aea <_entry+0x42> @ imm = #0xe
10000ada: 18c4         	adds	r4, r0, r3
10000adc: 3b04         	subs	r3, #0x4
10000ade: f854 4c08    	ldr	r4, [r4, #-8]
10000ae2: 42a1         	cmp	r1, r4
10000ae4: dbf6         	blt	0x10000ad4 <_entry+0x2c> @ imm = #-0x14
10000ae6: 4403         	add	r3, r0
10000ae8: e000         	b	0x10000aec <_entry+0x44> @ imm = #0x0
10000aea: 4603         	mov	r3, r0
10000aec: 6019         	str	r1, [r3]
10000aee: 3204         	adds	r2, #0x4
10000af0: f10e 0e04    	add.w	lr, lr, #0x4
10000af4: 4562         	cmp	r2, r12
10000af6: d1e8         	bne	0x10000aca <_entry+0x22> @ imm = #-0x30
10000af8: bdd0         	pop	{r4, r6, r7, pc}
10000afa: e8bd 40d0    	pop.w	{r4, r6, r7, lr}
10000afe: f7ff bba7    	b.w	0x10000250 <core::slice::sort::unstable::ipnsort::h6ad55a7d1a61fa1d> @ imm = #-0x8b2
10000b02: 4802         	ldr	r0, [pc, #0x8]          @ 0x10000b0c <_entry+0x64>
10000b04: 4478         	add	r0, pc
10000b06: f000 f864    	bl	0x10000bd2 <core::panicking::panic_const::panic_const_div_by_zero::h68898a40cd2b15cb> @ imm = #0xc8
10000b0a: bf00         	nop

10000b0c <$d>:
10000b0c: f8 f4 ff 0f  	.word	0x0ffff4f8
```

felix91gr: @adfernandes so then if I understood correctly... the zero check isn't being elided at all, right? And... is that the case for all optimization conditions?

I ask because if it were being elided under some condition, such that the program no longer aborted execution after it divided by zero, I think that would be a bug in either `rustc`'s codegen step or in LLVM.

adfernandes: @felix91gr, you are correct, the zero-check is **not** being elided, even under `opt=s` conditions.

Because of how LLVM handles the `udiv` and `sdiv` LLVM-IR instructions, these two LLVM-IR instructions _can_ produce undefined behaviour.

However, `rustc` is _explicitly_ adding a "zero-check" for the divisor. It would be a bug for **any** LLVM optimization or code-generation pass to elide the check, because doing so would _explicitly_ introduce the _possibility_ of undefined behavior (in the formal, LLVM sense).

The only way the check can be elided is if the optimizer can prove equivalence, such as inlining the function with known or constant values.

🥳

felix91gr: Wohoo, hell yeah! 🚀
(Thanks for the explanation n.n)

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,44 @@ Expressions
          }
 
          fn with_base(_: &Base) { ... }
+
+.. guideline:: Do not divide by 0
+   :id: gui_kMbiWbn8Z6g5
+   :category: Mandatory
```
PLeVasseur, line 87:
```suggestion
   :category: mandatory
```

Should be lowercase, see [here](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/blob/303fa8982caf4d7156712af0a6912841a81fe0d2/src/conf.py#L104-L109) in `conf.py`.

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,44 @@ Expressions
...
+   :status: draft
+   :release: latest
+   :fls: fls_Q9dhNiICGIfr
+   :decidability: Undecidable
```
PLeVasseur, line 91:
```suggestion
   :decidability: undecidable
```

Should be lowercase, see [here](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/blob/303fa8982caf4d7156712af0a6912841a81fe0d2/src/conf.py#L111-L114) in `conf.py`.

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,44 @@ Expressions
...
+   :scope: System
```
PLeVasseur, line 92:
```suggestion
   :scope: system
```

Should be lowercase, see [here](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/blob/303fa8982caf4d7156712af0a6912841a81fe0d2/src/conf.py#L116-L120) in `conf.py`.

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,44 @@ Expressions
...
+
+         let x = 0;
+         let x = 5 / x;
+
+   .. compliant_example::
+      :id: compl_ex_Ri9pP5Ch3kbb
+      :status: draft
+
+      There is no compliant way to perform integer division by zero. A checked division will prevent any
+      division by zero from happening. The programmer can then handle the returned Option.
```
PLeVasseur, line 120: [Check out](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/pull/127/files#diff-1078c969147388afdf0520a0bc985ca793fec1fbfa2b10c64b32df1dabdd8f8eR126) how we can use one of the nice bits of the coding guidelines extension to link to the Rust standard library.
```suggestion
      division by zero from happening. The programmer can then handle the returned :std:``std::option::Option``.
```
(This was shamelessly stolen from the FLS extension)

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,44 @@ Expressions
          }
 
          fn with_base(_: &Base) { ... }
+
+.. guideline:: Do not divide by 0
```
PLeVasseur, line 85: Asking a bit of a practical question to folks. Does the combination of:
* `mandatory`
* `undecidable`

put too great of a burden on reviewers / auditors or is this "just the way it is and should be" given the nature of writing safety-critical code?

Tagging @AlexCeleste and @rcseacord for their thoughts as well.

PLeVasseur (changes requested):
Thanks for pulling this together @vapdrs! I left a few suggestions based on how `conf.py` works together with Sphinx Needs as well as a way to use our coding guidelines extension. Could you take a look?

src/coding-guidelines/expressions.rst
felix91gr: Well, "do not divide by zero" is undecidable in the sense that statically knowing if the `b` in `a / b` is `0` is undecidable.

However, the guideline is not impossible to comply with since the user themselves can check (`if b != 0` etc) but also because there are APIs that we can use for this, like the `NonZero` types and the `checked_div` and `checked_rem` functions.

PLeVasseur: > However, the guideline is not impossible to comply with since the user themselves can check

Definitely, I agree. Perhaps this is an instance of where we need to make use of the **Exceptions** section to list the sorts of suggestions you made. What do you two think?

> like the `NonZero<T>`

TIL about [`NonZero<T>`](https://doc.rust-lang.org/stable/std/num/struct.NonZero.html). Super cool.

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,54 @@ Expressions
...
+   :fls: fls_Q9dhNiICGIfr
+   :decidability: undecidable
+   :scope: system
+   :tags: numerics
+
+   This guideline applies when unsigned integer or two’s complement division is performed during the
+   evaluation of an `ArithmeticExpression
+   <https://rust-lang.github.io/fls/expressions.html#arithmetic-expressions>`_.
+
+   Note that this includes the evaluation of a `RemainderExpression
```
AlexCeleste, line 99:
```suggestion
   This includes the evaluation of a `RemainderExpression
```
Since this is normative

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,54 @@ Expressions
...
+   <https://rust-lang.github.io/fls/expressions.html#syntax_remainderexpression>`_, which uses unsigned integer or two's
+   complement division.
+
+   This rule does not apply to evaluation of a :std:`core::ops::Div` trait on types other than `integer
+   types <https://rust-lang.github.io/fls/types-and-traits.html#integer-types>`_. The use of
+   :std:`std::num::NonZero` or is therefore a recommended way to avoid the undecidability of this
```
AlexCeleste, line 105: "or" seems to be a missing fragment here?

Should this be "as the divisor", or "or an equivalent", or something like that?

src/coding-guidelines/expressions.rst
```diff
@@ -83,6 +83,58 @@ Expressions
...
+   .. compliant_example::
+      :id: compl_ex_Ri9pP5Ch3kbb
+      :status: draft
+
+      There is no compliant way to perform integer division by zero. A checked division will prevent any
+      division by zero from happening. The programmer can then handle the returned :std:``std::option::Option``.
+
+      .. code-block:: rust
+
+         let x = 5u8.checked_div(0);
```
AlexCeleste, line 135: example maybe would look prettier with a similar `match` to the example in the unchecked-operators PR?

src/coding-guidelines/expressions.rst
vapdrs: Indeed, it was missing the former. 86fe09a3ddfbda066d0841f8adcb2326aac3f498

src/coding-guidelines/expressions.rst
```diff
@@ -83,6 +83,61 @@ Expressions
...
+      There is no compliant way to perform integer division by zero. A checked division will prevent any
+      division by zero from happening. The programmer can then handle the returned :std:``std::option::Option``.
+
+      .. code-block:: rust
+
+         let result = match 5u8.checked_div(0) {
+            None => 0
+            Some(r) => r
+         };
+
```
rcseacord, line 139: I would like to add another compliant solution here that looks like this:

Another compliant solution is to avoid division by zero:

.. code-block:: rust
 
if (x != 0) {
         let x = 5 / x;
}

felix91gr: This is a very reasonable way to comply as well.

If the blocks are small enough, I'm pretty sure both LLVM and GCC are able to optimize away the zero-division-abort that would otherwise be present (by way of value range analysis on the divisor).

vapdrs: I agree with your analysis @felix91gr, however I will point out that this Guideline is marked as "Undecidable" already, and follows the CERT style. As soon as you have that designation it is up to the quality of the static analysis tool's implementation as to whether it can or should do anything. The counterpart guideline in #136 is the decidable one.

I also do agree that,

```rust
if (x != 0) {
   let x = 5 / x;
}
else {
   let x = 0;
}
```

is compliant. However I'm not sure I want to promote that in our examples section, as that makes clippy developers sweat

felix91gr: You're totally right. This guideline is CERT-style, so it's about defects rather than subsets. 100% right.

src/coding-guidelines/expressions.rst
```diff
@@ -83,6 +83,61 @@ Expressions
...
 
+.. guideline:: Do not divide by 0
+   :id: gui_kMbiWbn8Z6g5
+   :category: required
+   :status: draft
+   :release: latest
+   :fls: fls_Q9dhNiICGIfr
+   :decidability: undecidable
+   :scope: system
+   :tags: numerics
```
felix91gr, line 94: I think we want the `defect` tag here

felix91gr (changes requested):

felix91gr:
I haven't been able to give the examples a proper re-read just yet. Tis' been a low-battery week. Will do my best to have final feedback on those (also regarding #136)

src/coding-guidelines/expressions.rst
vapdrs: Resolved in bbae26b

src/coding-guidelines/expressions.rst
vapdrs: After some more thought, I added the original example in 10efe3a. I did caveat it with a cautionary note in the description.

felix91gr:
@vapdrs thanks! I was gonna ping you here as well but you just did the thing :heart:

felix91gr:
This one is done, as far as I'm concerned. Much like for #136, thank you @vapdrs for your work and for your patience ^^

I'll add it to the merge queue right away :)

PLeVasseur (approved):
Thanks for seeing this through @vapdrs!

felix91gr (approved):

Issue #20, opened 2025-03-26
std::i32::MIN % -1 and related


rcseacord:
do you want to a rule that says used check remainder or make sure that your inputs are not std::i32::MIN % -1
probably std::i32::MIN / -1 and divide by zero as well divide by zero as well

```
use std::io;
fn main() {
    // Define `i` as the minimum value for a 32-bit signed integer
    let i = std::i32::MIN;
    // Prompt the user for input
    println!("Please enter an integer value for j:");
    // Read input from stdin
    let mut input = String::new();
    io::stdin().read_line(&mut input)
        .expect("Failed to read line");
    // Parse the input as an integer
    let j: i32 = match input.trim().parse() {
        Ok(num) => num,
        Err(_) => {
            eprintln!("Error: Please provide a valid integer for j.");
            std::process::exit(1);
        }
    };
    // Calculate and print the modulus
    println!("i % j = {}", i % j);
}
```

```
Exited with status 101
Standard Error
   Compiling playground v0.0.1 (/playground)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.68s
     Running `target/debug/playground`

thread 'main' panicked at src/main.rs:25:28:
attempt to calculate the remainder with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Standard Output
Please enter an integer value for j:  -1
```

Lot's of devs don't know about std::i32::MIN % -1, because mathematically it should be zero but faults on Intel processors.

If I ran this on a processor that doesn't fault, would it compile differently?

Compliant solution:

```
fn main() {
    let i = std::i32::MIN;
    let j = -1;
    match i.checked_rem(j) {
        Some(result) => println!("i % j = {}", result),
        None => println!("i % j = 0"), // This is mathematically correct
    }
}
```

x0rw:
Hey Robert, for safety-critical systems, unexpected runtime failures are unacceptable. Enforcing these checks aligns with guidelines like MISRA and ensures deterministic behavior, so the use of **checked arithmetic functions** should be **mandatory**. However, I think for a safety-critical guideline to be general, there should be a section about when **unchecked operations** can be used (in performance-critical real-time systems), along with guarantees (formal and manual verifications).

PLeVasseur:
> If I ran this on a processor that doesn't fault, would it compile differently?

This is a good question I don't know the answer to -- anyone know?

x0rw:
[...]

PR #136, opened 2025-06-24
[Guideline] Add safe division guideline to avoid dividing by zero

vapdrs:
Add a new guidlelines disallowing ArithmeticExpressions to use integer divisors. This is the "MISRA" style counterpart to #132

### DIFF
--- a/src/coding-guidelines/expressions.rst
+++ b/src/coding-guidelines/expressions.rst
@@ -108,8 +108,9 @@ Expressions
       * To create divisors using :std:`std::num::NonZero`, which then allows the programmer to perform those
         operations knowing that their divisor is not zero.
 
-      Since the compiler can assume the value of :std:`std::num::NonZero` is not zero, the check for zero
-      while dividing with standard integer types can be elided.
+      **Note:** since the compiler can assume the value of a :std:`std::num::NonZero`
+      variable to not be zero, checks for zero when dividing by it can be elided in the
+      final binary, increasing overall performance beyond what normal division can have.
 
    .. non_compliant_example::
       :id: non_compl_ex_0XeioBrgfh5z
@@ -128,7 +129,7 @@ Expressions
       :id: compl_ex_k1CD6xoZxhXb
       :status: draft
 
-      There is no compliant way to divide with an integer type, however here the developer explicitly:
+      There is no compliant way to divide with an integer type. Here, instead, the developer explicitly:
 
       * Uses a checked division function, which ensures a zero divisor is handled separately, and
       * Creates a divisor using :std:`std::num::NonZero`, which outsources the check for zero to the


vapdrs:
The `check_typos` job seems to have encountered a false positive looking at one of the ids

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,91 @@ Expressions
...
+   :category: required
+   :status: draft
+   :release: latest
+   :fls: fls_Q9dhNiICGIfr
+   :decidability: decidable
+   :scope: module
+   :tags: numerics
+
+   This guideline applies when an ArithmeticExpression is used with operands whose types evaluates integer or
+   floating-point types.
```
felix91gr, line 96: You know, I'm not sure floating-point types fit here.

* One one hand, the `f32` and other such types are defined to be [compliant with a specific IEEE standard](https://doc.rust-lang.org/std/primitive.f32.html).
* On the other hand, neither looking at that type's methods nor doing [a search for `checked_` methods in stdlib](https://doc.rust-lang.org/std/index.html?search=checked) I've been able to find anything like the safe/checked operations we have for integer types. I suspect they do not exist for floats.

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,91 @@ Expressions
...
+
+   .. rationale::
+      :id: rat_vLFlPWSCHRje
+      :status: draft
+
+      The built-in semantics for these expressions can result in panics, or silent wraparound when overflow,
+      underflow, or division by zero occurs. It is better to explicitly declare what should happen during
+      these events with checked arithmetic functions.
```
felix91gr, line 104: This seems perfect to me :ok_hand:

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,91 @@ Expressions
...
+      :status: draft
+
+      When the addition is performed, the overflow will be detected and the author will need to explicitly
+      deal with the Option value returned.
+
+      .. code-block:: rust
+
+         let y = 135u8
+         let x = 200u8.checked_add(y);
+
```
felix91gr, line 171: [There are many `unchecked` functions in `stdlib`](https://doc.rust-lang.org/std/index.html?search=unchecked). I reckon we might want to add a rule against all of them? I'm not sure. Some of them might be reasonable for SC contexts.

Regardless: just like for the rule above, the `unchecked_` arithmetic functions mentioned here only exist for integer types. Floats do have a `to_int_unchecked` method, though.

felix91gr:
Resolves #20

src/coding-guidelines/expressions.rst
rcseacord:
```suggestion
   This guideline applies when an `ArithmeticExpression` is used with operands of integer or
   floating-point type.
```

rcseacord: I think I agree I would separate integer from floating point types.

src/coding-guidelines/expressions.rst
rcseacord:
```suggestion
   .. rationale::
      :id: rat_vLFlPWSCHRje
      :status: draft

      The built-in semantics for these expressions can result in panics, or silent wraparound when overflow,
      underflow, or division by zero errors. 
      Explicitly declare what should happen during using checked arithmetic functions.
```

felix91gr: > Underflow is only a condition that occurs during floating point operations. Integers can only overflow. At least that's true in C.

I think that's true, but it's only true regarding the name – Underflow does occur in Rust, in the sense of attempting to go below the minimum value, but (I've just come to realize) [we call it Overflow](https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html#overflow) just the same.

rcseacord: Underflow is typically a floating point property.  Floating point is really weird though because almost nothing is an error if you anticipate it.

felix91gr: > Floating point is really weird though because almost nothing is an error if you anticipate it.

I have so many questions. Though floating point handling and its semantics are outside the topic of this PR. I'd like to ask you about floats in the context of SC. Are you perchance in the `rust-lang` Zulip?

vapdrs: Very interesting, you are right that Rust just calls all of it overflow for integers. Removed mentions of underflow in 8d2392d4a942207d325a842863c6c8e4f9d5756f

src/coding-guidelines/expressions.rst
vapdrs: You are correct, the float point types don't fit here. Removed, and resolved formatting/grammar in 8d2392d4a942207d325a842863c6c8e4f9d5756f

felix91gr:
Resolves #20

src/coding-guidelines/expressions.rst
PLeVasseur: TIL, interesting!

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,91 @@ Expressions
...
+      :id: compl_ex_k1CD6xoZxhXb
+      :status: draft
+
+      If an overflow occurs, it is explicit that the addition should wrap.
+
+      .. code-block:: rust
+
+         let y = 135u8
+         let x = 200u8.wrapping_add(y);
+
```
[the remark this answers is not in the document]
PLeVasseur: We are still at the point where these things can be discussed. If we'd like to have a number of compliant and non-compliant examples, this is possible likely in Sphinx Needs. Opened https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/issues/141 to track.

src/coding-guidelines/expressions.rst
PLeVasseur: @vapdrs or @felix91gr -- could you open a coding guideline issue to track this, if we think important but not addressing in this PR?

src/coding-guidelines/expressions.rst
PLeVasseur: Marking as resolved then!

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,101 @@ Expressions
...
+   :id: gui_mNEvznFjC3kG
+   :category: advisory
+   :status: draft
+   :release: latest
+   :fls: fls_Q9dhNiICGIfr
+   :decidability: decidable
+   :scope: module
+   :tags: numerics
+
+   This guideline applies to any call to the integer type methods that begin with `unchecked_`.
```
PLeVasseur, line 147: Two things:
1. I think we need [double back-ticks for code-style formatting](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup) in ReStructured Text.
2. It would be great to offer up an example of an `unchecked_` function for the reader of the guideline to read on for more. Seems the preferred way to do this is documented [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links).
```suggestion
   This guideline applies to any call to the integer type methods that begin with ``unchecked_``, such as ``core::primitive::u8::unchecked_add``_.
   
 .. _core::primitive::u8::unchecked_add: https://doc.rust-lang.org/std/primitive.u8.html#method.unchecked_add
```
(I didn't try to render the link above to `unchecked_add`, but reading the docs, this seems like it should work. Apologies if you have to fiddle a bit. 🙏 )

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,101 @@ Expressions
...
+
+   .. rationale::
+      :id: rat_7tF18FIwSYws
+      :status: draft
+
+      The semantics for these expressions can result in panics, or silent wraparound upon overflow or division
```
PLeVasseur, line 153: nit

Suggest either "wrap-around" or "wrap around". Them concatenated together somehow looks a bit off to me 😅

src/coding-guidelines/expressions.rst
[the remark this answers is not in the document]
felix91gr: @PLeVasseur I would much prefer it if we had multiple examples, one for each specific case. Whenever I've read guidelines, having "atomic" (ie indivisible / minimal) examples has been essential for me to understand them quickly and without needing to ask for further help. I think we can benefit a lot if each example addresses a single concern.

So yeah, I think it's best if we block until we have this capability. Sorry, @vapdrs, it might be a pain to wait.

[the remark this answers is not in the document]
vapdrs: No worries here. I don't mind waiting.

src/coding-guidelines/expressions.rst
vapdrs: Agreed, resolved in 9dc042f

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,104 @@ Expressions
...
+   :tags: numerics
+
+   This guideline applies to any call to the integer type methods that begin with ``unchecked_``, such as
+   `core::primitive::u8::unchecked_add <https://doc.rust-lang.org/std/primitive.u8.html#method.unchecked_add>`_.
+
+   .. rationale::
+      :id: rat_7tF18FIwSYws
+      :status: draft
+
+      The semantics for these expressions can result in panics, or silent wrap around upon overflow or division
```
luca3s, line 156: This was probably just copied from above, but it is wrong. The unchecked functions lead to UB when overflow happens, so they are more dangerous than the regular arithmetic expressions.

vapdrs: Thank you for pointing this out, resolved in 2fb316f

luca3s: You forgot the next sentence. There it also says panic.

src/coding-guidelines/expressions.rst
```diff
@@ -81,3 +81,104 @@ Expressions
...
+      :id: rat_7tF18FIwSYws
+      :status: draft
+
+      The semantics for these expressions can result in undefined behavior in situations where an equivalent
+      checked operation would return ``None``. It is recommended to explicitly declare what should happen
+      during these events with checked arithmetic functions.
+
+      In a particularly performance sensitive critical section of the code it may be necessary to use the
+      unchecked methods in tandem with assurances that the arguments will never meet the undefined behavior
+      conditions.
```
pellico, line 162: It seems to me that there is no need for this rule.  
To use unchecked arithmetic, you need an `unsafe` block, as with any function that can lead to undefined behavior.  
For `unsafe` blocks, we can make it mandatory to provide a justification (see the Clippy rule `undocumented_unsafe_blocks`).  
I expect that developers will not write `unsafe` code unless it is necessary.  
Therefore, this rule seems equivalent to _"don't write unsafe code"_.

Am I missing something?

adfernandes:
As @rcseacord mentioned, using checked method calls everywhere is not very ergonomic. 

It also can be too slow for many embedded applications.

In `no_std` Rust, panics may be `abort` and not `unwind`, so you cannot assume that you have a panic handler. Even in `std` Rust, panics are not guaranteed to be recoverable.

On Arm Cortex-M targets, integer division by zero triggers a `UsageFault` interrupt. The standard behaviour is to us **un**-checked arithmetic in time-sensitive applications, with the user "proving" to what extent they can that division by zero never happens. If it **does**, that triggers the interrupt, but does not necessarily terminate the program.

I'm not saying that's the best strategy, but that's what's done today.

src/coding-guidelines/expressions.rst
PLeVasseur: Bump to surface for @vapdrs

iglesias:
On the one hand, I think the potential issue with using checked operations making it not very ergonomic is resolved by using ("strong") types and operator overloading. On the other hand, the issue about performance is different, I think there's a trade-off between performance and safety in that case.

I think it sounds good that there is one or more guidelines for checked and unchecked operations, division by zero, possibly etc. And also think that it can be OK if any of those are categorized compulsory or required.
Consequently, it would be fair if a code analysis tool would flag some code as non-compliant according to any of those guidelines. And, since there's a trade-off, I think it would also be OK that developers choose to deviate and silent such diagnostic. Hopefully, the guideline and corresponding diagnostic from the code analysis tool would have prompted the developers to check and make sure that even when for some reason (such as deficiency of the range analysis in the tool, or undecidability due to e.g. user input) the code is flagged, it is indeed safe.

src/coding-guidelines/expressions.rst
vapdrs: Apologies for the late response, just got back from vacation.

I agree with you on this matter. I had missed that these function were marked as `unsafe`, and I indeed think the advisory rule is unnecessary.

src/coding-guidelines/expressions.rst
```diff
@@ -82,6 +82,58 @@ Expressions
...
+
+         let x = 0;
+         let result = match 5u32.checked_div(x) {
+           None => 0
+           Some(r) => r
+         }
+         if let Some(divisor) = match NonZero::<u32>::new(x) {
+           let result = 5 / divisor;
+         }
+
```
felix91gr, line 136: @vapdrs Alex had a brilliant idea. Instead of making you wait for the multiple example functionality to be added to our pipeline, you could add all the non-compliant examples in one code block and all the compliant examples in another code block. Then whenever we are able to add multiple examples, they can be split into their separate blocks.

How does that sound? :)

felix91gr (changes requested):
It looks good to me, I think. We'll get the multiple example functionality later; for now, just add all the compliant examples under one block and all the non compliant ones under the other block. Please feel free to put each example's description in a comment above it (inside the code block) or however it feels right :)

src/coding-guidelines/expressions.rst
```diff
@@ -82,6 +82,59 @@ Expressions
...
 
+.. guideline:: Do not use integer type as divisor
+   :id: gui_7y0GAMmtMhch
+   :category: advisory
+   :status: draft
+   :release: latest
+   :fls: fls_Q9dhNiICGIfr
+   :decidability: decidable
+   :scope: module
+   :tags: numerics
```
felix91gr, line 93: I think we want to add here the tag `subset` and perhaps `reduce-human-error` as well

felix91gr (changes requested):

felix91gr:
Damn, I thought the PR would fix itself after a non-conflicting change to `main` was made. But alas, `git`'s commits are snapshots of the directory, and not changes. It seems like `git` believes your copy of the spec lock file to be part of your changes, or something.

Because when looking at this branch, [the spec lock_file is outdated](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/blob/557bb5ca7e49a0c9d62615263b4124e3279ced9e/src/spec.lock).

@vapdrs can I ask you to re-merge the latest commits from `main` into your branch? Sorry for the bother.

With those changes, it should build and I'll be able to give a reasonable final review :3

src/coding-guidelines/expressions.rst
```diff
@@ -82,6 +82,59 @@ Expressions
...
+         let x = 0;
+         let y = 5 / x;
+         let x = 5 % x;
+
+   .. compliant_example::
+      :id: compl_ex_k1CD6xoZxhXb
+      :status: draft
+
+      The developer must explicitly indicate the intended behavior when a division by zero occurs, or use a
+      type for which it is invalid to have a value of zero.
```
felix91gr, line 125: Like in my first comment, I don't think the description here of "explicitly indicate the intended behavior" makes sense now that the PR is focusing uniquely on the division-by-zero scenario.

I'd also change the second point, since it's not that a `NonZero<T>` with an inner value of zero is invalid.

It's more like it's impossible to have a `NonZero<T>` with an inner value of zero - unless the user constructs one using `unsafe` and uses the `unchecked_` constructor, which like everything regarding `unsafe`, probably merits its own entire guideline series. Though in that case, if the user _does_ construct a `NonZero<T>` using a `T` of value zero through the `unchecked_` method, that's Undefined Behavior.

Anyway. I'm thinking of something like this:

```
The developer must explicitly:

* Use checked division functions, which make sure a zero divisor is handled separately, or
* To first create divisors using :std:`std::num::NonZero`, which let the programmer prove the divisor is not zero. 
```

felix91gr (changes requested):
I hope I'm not boring you :heart_hands: 

This is mostly details - the content and structure itself all look good to me.

felix91gr:
Indeed! Dividing by `NonZero<T>` produces a faster assembly than `/` and `checked_div`: https://play.rust-lang.org/?version=stable&mode=release&edition=2024&gist=c8fc94e2335fcba5921220be63542786 (instead of Run, select "Show LLVM IR"

src/coding-guidelines/expressions.rst
vapdrs: I agree, I rephrased the suggestion though

src/coding-guidelines/expressions.rst
```diff
@@ -82,6 +82,69 @@ Expressions
...
+
+   .. compliant_example::
+      :id: compl_ex_k1CD6xoZxhXb
+      :status: draft
+
+      There is no compliant way to divide with an integer type, however here the developer explicitly:
+
+      * Uses a checked division function, which make sure a zero divisor is handled separately, and
+      * Creates a divisor using :std:`std::num::NonZero`, which outsources the check for zero to the
+        construction of that struct.
```
felix91gr, line 135:
```suggestion
        construction of that struct. It's worth noting that such a divisor can be used multiple times after it's been created, whilst keeping the guarantee that such divisions will be safe.
```

I'd add this and extend the example to use the `divisor` more than once, to show the main difference in how the `checked_` APIs and the `NonZero<T>` types can be used.

src/coding-guidelines/expressions.rst
```diff
@@ -82,6 +82,69 @@ Expressions
...
+      :id: non_compl_ex_0XeioBrgfh5z
+      :status: draft
+
+      When either the division or remainder are performed, the right operand is evaluated to zero and the
+      program panics.
+
+      .. code-block:: rust
+
+         let x = 0;
+         let y = 5 / x;
```
felix91gr, line 124:
```suggestion
         let y = 5 / x; // This line will panic.
```

I think this kind of comment always helps a little bit.

src/coding-guidelines/expressions.rst
```diff
@@ -82,6 +82,69 @@ Expressions
...
+         let x = 5 % x;
```
felix91gr, line 125:
```suggestion
         let z = 5 % x; // This line would also panic.
```

Added the same kind of comment as above, plus changed the variable to `z` to keep things extra clear (by not using `x` anywhere but as a divisor)

src/coding-guidelines/expressions.rst
```diff
@@ -82,6 +82,69 @@ Expressions
...
+
+      The built-in semantics for these expressions can result in panics when division by zero occurs. It is
+      recommended to either:
+
+      * Use checked division functions, which ensure the programmer handles the case when the divisor is zero, or
+      * To create divisors using :std:`std::num::NonZero`, which then allows the programmer to perform those
+        operations knowing that their divisor is not zero.
+
+      Since the compiler can assume the value of :std:`std::num::NonZero` is not zero, the check for zero
+      while dividing with standard integer types can be elided.
```
felix91gr, line 112:
```suggestion
      **Note:** since the compiler can assume the value of a :std:`std::num::NonZero`
      variable to not be zero, checks for zero when dividing by it can be elided in the
      final binary, increasing overall performance beyond what normal division can have.
```

src/coding-guidelines/expressions.rst
```diff
@@ -82,6 +82,69 @@ Expressions
...
+
+         let x = 0;
+         let y = 5 / x; // This line will panic.
+         let z = 5 % x; // This line would also panic.
+
+   .. compliant_example::
+      :id: compl_ex_k1CD6xoZxhXb
+      :status: draft
+
+      There is no compliant way to divide with an integer type, however here the developer explicitly:
```
felix91gr, line 131:
```suggestion
      There is no compliant way to divide with an integer type. Here, instead, the developer explicitly:
```

felix91gr (changes requested):
Beyond these changes, I have nothing more to ask. The PR is done.

Great job, @vapdrs! And thank you for your patience :)

felix91gr (approved):

PLeVasseur (approved):
@vapdrs -- thanks for working with us as we reviewed ☺️

felix91gr:
⛵

PR #155, opened 2025-07-23
Add CNAME to deploy step

PLeVasseur:
So that the coding guidelines can be seen at coding-guidelines.arewesafetycriticalyet.org

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -27,6 +27,10 @@ jobs:
           ls build
           ls build/html
 
+      - name: Add CNAME to be able to be seen at coding-guidelines.arewesafetycriticalyet.org
+        run: |
+          echo "coding-guidelines.arewesafetycriticalyet.org" >> ./build/html/CNAME
+
       - name: Deploy to GitHub Pages
         uses: peaceiris/actions-gh-pages@v4
         with:


.github/workflows/deploy.yml
```diff
@@ -27,6 +27,10 @@ jobs:
           ls build
           ls build/html
 
+      - name: Add CNAME to be able to be seen at coding-guidelines.arewesafetycriticalyet.org
+        run: |
+          echo "coding-guidelines.arewesafetycriticalyet.org" >> ./build/html/CNAME
```
felix91gr, line 32: We've talked about this on our Zulip DMs and since this file doesn't persist, this command should be fine.

However, just to double down on our intent, we'd perhaps want it to have a single `>`, so as to:
- (Like right now) create the file and put this line on it if it doesn't exist,
- **Replace** the existing file and put this line on it if it does exist.

Basically:

`echo "coding-guidelines.arewesafetycriticalyet.org" > ./build/html/CNAME`

The difference is:
- `>>` creates or _appends_
- `>` creates or _replaces_

That way we don't depend on the non-persistence of that file. Otherwise this looks perfect to me c:

felix91gr (changes requested):
I like it!

Added a tiny suggestion to avoid depending on the fact that this file doesn't persist :3

.github/workflows/deploy.yml
PLeVasseur: Makes sense! Updated

felix91gr (approved):

PR #171, opened 2025-08-08
Bump spec lock file

felix91gr:
What it says in the title :)


felix91gr:
@PLeVasseur this is what you meant, right? :)

PR #172, opened 2025-08-11
Bump spec lock file

felix91gr:
CC @PLeVasseur I'm syncing the spec lock file again to upstream; #132 and #136 can't currently build because of it x3

I think I did it alright? Lmk if I should change anything

### DIFF
--- a/src/spec.lock
+++ b/src/spec.lock
@@ -245,2142 +245,227 @@
             "title": "Associated Items"
         },
         {
-            "informational": false,
-            "link": "attributes.html",
-            "sections": [
-                {
-                    "id": "fls_gvwd0kf72jt",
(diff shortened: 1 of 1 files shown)


felix91gr:
@PLeVasseur I used the output of https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/actions/runs/16881087242/job/47816848850?pr=136#step:4:187 as a reference. It requires changes to 0 guidelines, so I assumed it would be okay to just run the command of step 6.

Though lemme just quickly check that `main` alone also gives that output. It might be the case that another change to the FLS has happened in the meantime which the Build process in CI hasn't seen yet.

felix91gr:
:boat:

PR #175, opened 2025-08-18
 [Guideline] Do not invoke undefined behavior

AchimKriso:
I felt motivated to write some rules about unsafety in Rust.

This rule (or a variant thereof) seems inevitable, since most coding standards I'm aware of have a similar rule:

* CERT: [MSC15-C. Do not depend on undefined behavior](https://wiki.sei.cmu.edu/confluence/display/c/MSC15-C.+Do+not+depend+on+undefined+behavior)
* MISRA C 2025 Rule 1.3 and MISRA C++ 2023 Rule 4.1.3: There shall be no occurrence of undefined or critical unspecified behaviour

PR #176, opened 2025-08-18
[Guideline] Add guideline requiring wrapping unsafe code in safe abstractions

AchimKriso:
Like #175, this is a low hanging fruit of a guideline.

While there aren't analogous guidelines for CERT or MISRA (as far as I know) this is mostly due to the fact that C or C++ do not have the same safe-unsafe split.

Still, without this rule the whole concept of safe Rust becomes meaningless.

### DIFF
--- a/src/coding-guidelines/unsafety.rst
+++ b/src/coding-guidelines/unsafety.rst
@@ -43,8 +43,8 @@ Unsafety
 
         .. code-block:: rust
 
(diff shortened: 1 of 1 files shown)


AchimKriso:
Hi @PLeVasseur, I've seen the comment, I'll try to get to it this or one of the next days.

PR #184, opened 2025-08-25
Allow only a single selection for Category

felix91gr:
Cleaning up the Issue Template :)


PLeVasseur (approved):
@felix91gr -- Thanks for finding this issue!

PR #191, opened 2025-08-26
Github action to update FLS every 24 hours

### LINKED ISSUE #173: [Automation] Update the FLS spec lock_file whenever it's trivial to do
Can we have some sort of `cron` job in Github?

Anywho.

Here's the issue at hand: every so often, the FLS changes and we need to update the spec's lock_file. Sometimes, this change affects certain guidelines, and for those cases some amount of non-trivial work is needed to update the project accordingly.

However, most of the time such changes in the FLS are quite trivial and can be resolved with a simple command:

`./make.py --update-spec-lock-file`

Right now, whenever such changes in the FLS occur, all the active Pull Requests stop being able to build, as seen [here](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/actions/runs/16881081434/job/47816828942?pr=132#step:4:186). And in order to build once again, an auxiliary PR like #172 needs to be opened and merged.

That's a bit of a waste. It would be great if such changes could be made automatically - ideally, with a cadence of a day or something similar. Plus an event-triggered change whenever a Pull Request is opened.

How does that sound? 🙂

plaindocs:
- Runs at 00:05 (or manually)
- Opens a PR if there are changes.
- Fixes #173

I did look at Dependabot, but I don't think it quite works for this scenario.

### DIFF
--- a/scripts/test_update_fls.sh
+++ b/scripts/test_update_fls.sh
@@ -14,6 +14,7 @@ else
     if [ $ZEROAFFECTED -eq "0" ]; then
       ./make.py --update-spec-lock-file
       rm test.txt
+      echo 'true' >> "$CAN_AUTOMATICALLY_UPDATE"
       exit 1 # Can be updated automatically
     else
       ./make.py --update-spec-lock-file


.github/workflows/update-fls.yaml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+    update-fls:
+      runs-on: ubuntu-latest
+      steps:
+        - name: Check out the repo
+          uses: actions/checkout@v4
+        - name: Install uv
+          uses: astral-sh/setup-uv@v6
+        - name: Update FLS
+          run: "./make.py --update-spec-lock-file"
```
felix91gr, line 15: The one thing that worries me is whether this would open a PR if there are affected guidelines: https://github.com/rustfoundation/safety-critical-rust-coding-guidelines?tab=readme-ov-file#if-you-need-to-audit-the-difference

I'm assuming that perhaps this is skipping next step depending on the return value of the command? And that in turn handles this edge case?

felix91gr: @plaindocs

PLeVasseur (changes requested):

.github/workflows/update-fls.yaml
plaindocs: I'll poke at this with a view to checking the normal build first, but not till next week.

plaindocs:
I've not looked into the Python version build error.

builder/build_cli.py
```diff
@@ -49,7 +49,7 @@ def build_docs(
 
     dest = root / "build"
 
-    args = ["-b", builder, "-d", dest / "doctrees"]
+    args = ["-b", builder, "-d", dest / "doctrees", "-wtest.txt"]
```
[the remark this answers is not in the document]
felix91gr: > This needs moving out into a flagged parameter so it doesn't happen with every build.

What would be the consequence if it happened on every build? @plaindocs

PLeVasseur:
Hi @plaindocs -- I tried my hand at this in https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/pull/382

Let's see how it runs tonight. If it's FUBAR'ed I may still ask for your help with it 🙏 

Will close since this issue was tackled by that PR, but feel free to ping me about it here or on Zulip.

PR #208, opened 2025-10-07
Fix typo in src/coding-guidelines/macros.rst (second attempt, now in-tree)

felix91gr:
CC @plaindocs so that you can cherrypick


felix91gr:
@PLeVasseur fixed the stalemate so this PR is now redundant ^^

PR #232, opened 2025-11-26
revised divide by zero rule

rcseacord:
Updated the divide by zero rule

### DIFF
--- a/src/coding-guidelines/expressions.rst
+++ b/src/coding-guidelines/expressions.rst
@@ -135,7 +135,7 @@ Expressions
       :id: compl_ex_k1CD6xoZxhXb
       :status: draft
 
-      Checked division prevents division by zero from occuring.
+      Checked division prevents division by zero from occurring.
       The programmer can then handle the returned :std:`std::option::Option`.
       Using checked division and remainder is particularly important in the signed integer case,
       where arithmetic overflow can also occur when dividing the minimum representable value by -1.


rcseacord:
@felix91gr  @PLeVasseur  this rule is ready for review.

I'm considering removing the duplicate compliant solutions from the required rule and referring to the advisory rule for these to eliminate redundancy.

This will hurt readability some, so perhaps we don't want to do this... particularly if we view this as a reference document.

src/coding-guidelines/expressions.rst
```diff
@@ -129,24 +136,42 @@ Expressions
...
+         };
+
+   .. compliant_example::
+      :id: compl_ex_k1CD6xoZxhXc
+      :status: draft
 
-      * Uses a checked division function, which ensures a zero divisor is handled separately, and
-      * Creates a divisor using :std:`std::num::NonZero`, which outsources the check for zero to the
-        construction of that struct. It's worth noting that such a divisor can be used multiple times after it's been created, whilst keeping the guarantee that such divisions will be safe.
+      This compliant solution creates a divisor using :std:`std::num::NonZero`,
```
felix91gr:
```suggestion
      This compliant solution creates a divisor using :std:`std::num::NonZero`.
```

A small typo

src/coding-guidelines/expressions.rst
```diff
@@ -129,24 +136,42 @@ Expressions
...
+      :std:`std::num::NonZero` is a wrapper around primitive integer types that guarantees the contained value is never zero.
+      :std:`std::num::NonZero` creates a new type that represents a value that is known not to be zero,
+      ensuring that functions operating on these types can safely assume that zero is not a possible input.
```
felix91gr:
```suggestion
      :std:`std::num::NonZero::new` creates a new binding that represents a value that is known not to be zero.
      This in turn ensures that functions operating on its value can safely assume that they are not being given zero their input. 
```

Making the link more precise, and the wording as well. I've cut the composite sentence into its two components; I believe the paragraph is a bit more readable this way. Lemme know what you think @rcseacord

src/coding-guidelines/expressions.rst
```diff
@@ -129,24 +136,42 @@ Expressions
...
+      :status: draft
 
-      * Uses a checked division function, which ensures a zero divisor is handled separately, and
-      * Creates a divisor using :std:`std::num::NonZero`, which outsources the check for zero to the
-        construction of that struct. It's worth noting that such a divisor can be used multiple times after it's been created, whilst keeping the guarantee that such divisions will be safe.
+      This compliant solution creates a divisor using :std:`std::num::NonZero`,
+      :std:`std::num::NonZero` is a wrapper around primitive integer types that guarantees the contained value is never zero.
+      :std:`std::num::NonZero` creates a new type that represents a value that is known not to be zero,
+      ensuring that functions operating on these types can safely assume that zero is not a possible input. 
+      The test for arithmetic overflow in this compliant example is unnecessary because ``divisor`` is an unsigned integer type.
```
felix91gr:
```suggestion

      Note that the test for arithmetic overflow in this compliant example (which is a possibility for non-zero negative divisors, namely, ``-1``) is unnecessary because ``divisor`` is an unsigned integer type.
```

I'm putting this on a separate paragraph because it's a separate idea. Lemme know what you think, @rcseacord

PLeVasseur (approved):
Hey thanks @rcseacord for revising based on feedback. LGTM, let's merge

PR #236, opened 2025-12-02
Update spec.lock to fix CI build

MarcosBorgesPhD:
This PR updates `src/spec.lock` to match the current FLS specification.


felix91gr:
@PLeVasseur I'm merging this so that the other PRs can continue building :)

PR #240, opened 2025-12-04
#235 [Coding Guideline]: Do not read uninitialized memory of any non-union type as a typed value

### LINKED ISSUE #235: [Coding Guideline]: Do not create values from uninitialized memory
### Chapter

Values

### Guideline Title

Do not create values from uninitialized memory except for union fields

### Category

Mandatory

### Status

Draft

### Release Begin

unclear

### Release End

latest

### FLS Paragraph ID

fls_6lg0oaaopc26

### Decidability

Undecidable

### Scope

System

### Tags

undefined-behavior

### Amplification

A program shall not create a value of any type from uninitialized memory, except when accessing a field of a union type, where such reads are explicitly defined to be permitted even if the bytes of that field are uninitialized.

It is prohibited to interpret uninitialized memory as a value of any Rust type (primitive, aggregate, reference, pointer, struct, enum, array, tuple, etc.) 

### Exception(s)

You can access a field of a union even when the backing bytes of that field are uninitialized provided that:

- The resulting value has an unspecified but well-defined bit pattern.
- Interpreting that value must still comply with the requirements of the accessed type (e.g., no invalid enum discriminants, no invalid pointer values, etc.).

For example, reading an uninitialized `u32` field of a union is allowed; reading an uninitialized `bool` field is disallowed because not all bit patterns are valid.

### Rationale

Rust’s memory model treats all types except unions as having an invariant that all bytes must be initialized before a value may be constructed. Reading uninitialized memory:

- creates undefined behavior for most types,
- may violate niche or discriminant validity,
- may create invalid pointer values,
- or may produce values that violate type invariants.

The sole exception is that unions work like C unions: any union field may be read, even if it was never written. The resulting bytes must, however, form a valid representation for the field’s type, which is not guaranteed if the union contains arbitrary data.

### Non-Compliant Example - Prose

This noncompliant code example creates a value from uninitialized memory via `assume_init`:

### Non-Compliant Example - Code

use std::mem::MaybeUninit;

let x: u32 = unsafe { MaybeUninit::uninit().assume_init() };  // UB

### Compliant Example - Prose

This compliant solution reads a union field:

### Compliant Example - Code

union U {
    x: u32,
    y: f32,
}

let u = U { x: 123 };       // write to one field
let f = unsafe { u.y };     // reading the other field is allowed

manhatsu:
Closes #235.


rcseacord:
I've made some improvement suggestions in a PR here:  https://github.com/manhatsu/safety-critical-rust-coding-guidelines/pull/1

PLeVasseur (commented):
Hi @manhatsu -- thank you for contributing. Please see the comment I left on how to generate a template.

workingjubilee (changes requested):
What @inkreasing says is correct. This description is insufficient to reflect the restrictions imposed by MIRI here:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=abb9da1c391902b21c03ed1d21767b58

Note this is not UB before line 15.

Bytes remain uninit until written. You may not read uninitialized bytes as any initialized type, period, not even if "all" bitpatterns are considered valid, because uninit is the 257th bitpattern for a byte, effectively: `0xUU`. By contrast, `u8` is `0x00` through `0xFF`, inclusive. We use `MaybeUninit<u8>` to indicate the final state is possible, and it is valid to read that value (well, from any allocation that has a byte in it, at least).

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,185 @@
...
+   :fls: fls_6lg0oaaopc26
+   :decidability: undecidable
+   :scope: system
+   :tags: undefined-behavior, unsafe
+
+   A program shall not create a value of any type from uninitialized memory,
+   except when accessing a field of a union type,
+   where such reads are explicitly defined to be permitted even if the bytes of that field are uninitialized.
+   It is prohibited to interpret uninitialized memory as a value of any Rust type such as a
+   primitive, aggregate, reference, pointer, struct, enum, array, or tuple.
```
workingjubilee: This definition does not consider composition:
```rust
use std::mem::MaybeUninit;

union Uninit32 {
    u: u32,
    i: i32,
    f: f32,
    void: (),
}

struct Newtype32(Uninit32);

fn main() {
    let x: Newtype32 = unsafe { MaybeUninit::uninit().assume_init() };
}
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=feae8c987fa0b2703533ce0ebf8b23ba

This passes miri because all the bytes in Newtype32 are defined by Uninit32, which is allowed to be uninitialized.

When bytes are read by `pointer::read`, `*ptr`, or `mem::transmute`, a "typed read" occurs. This asserts the bytes are valid *as that type*. `Uninit32` and `MaybeUninit` are the same thing here: unions with `()` as a possibility, which means they *must* be valid to read from a blob of uninitialized bytes within a valid allocation[^1]. Because `Newtype32` is entirely defined by `Uninit32`, it is also valid to read from uninitialized bytes: the `struct` wrapper does not impose a novel validity requirement. If this is a mandatory guideline, it should be more exacting about why.

[^1]: Note that this is likely a stronger requirement than the actual rules will be regarding union validity once final details of those are hashed out. I'm just giving an example that is very "in the clear".

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,185 @@
...
+   
+   **Exception:** You can access a field of a union even when the backing bytes of that field are uninitialized provided that:
+
+   - The resulting value has an unspecified but well-defined bit pattern.
```
workingjubilee: This operation is called "freeze" and it has not been admitted to the language yet. It may require a special operation, so it should not be assumed to be a property of every read operation.

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,185 @@
...
+   - Interpreting that value must still comply with the requirements of the accessed type
```
workingjubilee: What does "interpreting" mean here?

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,185 @@
...
+     (e.g., no invalid enum discriminants, no invalid pointer values, etc.).
+
+   For example, reading an uninitialized u32 field of a union is allowed;
+   reading an uninitialized bool field is disallowed because not all bit patterns are valid.
```
workingjubilee: It is not permitted to read a `bool` from anything but `0x00` and `0x01`, yes.

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,185 @@
...
+
+      This compliant example calles the ``write`` function to fully initialize low-level memory.
+
+      .. code-block:: rust
+
+         use std::mem::MaybeUninit;
+
+         let mut x = MaybeUninit::<u64>::uninit();
+         x.write(42);
+         let val = unsafe { x.assume_init() }; // OK — value was fully initialized
```
workingjubilee: Yes, this is correct.

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,185 @@
...
+         }
+
+        let u = U { bytes: [0xFF, 0xEE, 0xDD, 0xCC] };
+        let n = unsafe { u.n };   // OK — all bit patterns valid for u32
+
+   .. compliant_example::
+      :id: compl_ex_Ke869nSXuShV
+      :status: draft
+
+      This compliant example calles the ``write`` function to fully initialize low-level memory.
```
workingjubilee:
```suggestion
      This compliant example calls the ``write`` function to fully initialize low-level memory.
```

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,185 @@
...
+
+      Not all bit patterns are valid pointers for all operations (e.g., provenance rules).
+      You cannot create a pointer from unspecified bytes.
+      Even a raw pointer type (e.g., ``*const T``) has validity rules.
+
+      .. code-block:: rust
+
+         use std::mem::MaybeUninit;
+
+         let p: *const u32 = unsafe { MaybeUninit::uninit().assume_init() }; // UB
```
workingjubilee: Yes but not for the reasons you said.

To explain:
```rust
let bytes: [u8; 8] = core::from_fn(|_| random());
// this transmute is fine, actually
let ptr: *mut Type = unsafe { mem::transmute(bytes) };
// UB only triggers here:
let value = unsafe { *ptr };
```

`ptr` has no provenance. Not just because it was obtained from random bytes instead, but because `transmute` of int-to-ptr does not preserve provenance on the integer. So this is a no-provenance pointer. It is permitted to create this, but the read requires a provenance to be attached to be correct, so it's not.

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,185 @@
...
+
+      .. code-block:: rust
+
+         union U {
+             b: bool,
+             x: u8,
+         }
+
+         let u = U { x: 255 };        // 255 is not a valid bool representation
+         let b = unsafe { u.b };      // UB — invalid bool
```
workingjubilee: Correct.

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,185 @@
...
+             a: u8,
+             b: u32,
+         }
+
+         let mut buf = [0u8; std::mem::size_of::<S>()];
+         buf[0] = 10;
+         buf[1] = 20; // writing padding is fine
+
+         let p = buf.as_ptr() as *const S;
+         let s = unsafe { p.read_unaligned() }; // OK — all *fields* are initialized (padding doesn’t matter)
```
workingjubilee: I truly do not understand this example. All the bytes in the original buffer are initialized to 0, so of course reading them as a struct that permits `0x00..=0xFF` in every byte of every field is permitted?

The wording feels strange. It could be just "initialized bytes are valid to interpret as pattern bytes upon typed read".

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,185 @@
...
+      :status: draft
+
+      This noncompliant example creates a value of type ``u32`` from uninitialized memory via 
+      `assume_init <https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#method.assume_init>`_:
+
+      .. code-block:: rust
+
+         use std::mem::MaybeUninit;
+
+         let x: u32 = unsafe { MaybeUninit::uninit().assume_init() }; // UB
```
workingjubilee: You got this right, but the lesson needs to be extended elsewhere: `assume_init` and the read of a union field at its type are not really different operations in the semantics.

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,185 @@
...
+
+      Creating a reference from arbitrary or uninitialized bytes is always undefined behavior.
+      References must be valid, aligned, properly dereferenceable, and non-null.
+      Uninitialized memory cannot satisfy these invariants.
+
+      .. code-block:: rust
+
+         use std::mem::MaybeUninit;
+
+         let r: &u32 = unsafe { MaybeUninit::uninit().assume_init() }; // UB — invalid reference
```
workingjubilee: Yes, but I do not believe this has adequate explanatory power, so it should not be included in its current state: "arbitrary" is too vague. I recommend you do not include this example.

felix91gr (changes requested):
I hope this helps

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,263 @@
...
+   :id: gui_uyp3mCj77FS8
+   :category: required
+   :status: draft
+   :release: <TODO>
+   :fls: fls_6lg0oaaopc26
+   :decidability: undecidable
+   :scope: system
+   :tags: undefined-behavior, unsafe
+
+   Do not create a typed value from uninitialized memory.
```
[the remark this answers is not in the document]
felix91gr: You might also want to take a look at these pages:

1. The Rustonomicon on working with uninitialized memory from `unsafe` https://doc.rust-lang.org/nomicon/unchecked-uninit.html
2. The explanation of Initialization Invariant, examples and such, from the docs of [`MaybeUninit`](https://doc.rust-lang.org/core/mem/union.MaybeUninit.html#initialization-invariant).
3. The Language Reference on unions: https://doc.rust-lang.org/reference/items/unions.html

The first two contain most of what I think you'll need to complete this / these Guideline(s). I hope they help.

workingjubilee (commented):
Much better.

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,254 @@
...
+             a: u8,
+             b: u32,
+         }
+
+         fn make_s(a: u8) -> S {
+             let mut tmp = MaybeUninit::<S>::uninit();
+
+             unsafe {
+                 (*tmp.as_mut_ptr()).a = a;
+                 tmp.assume_init()
```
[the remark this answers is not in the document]
rcseacord: resolved

src/coding-guidelines/values.rst
rcseacord: resolved

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,263 @@
...
+     (e.g., ``enum`` discriminants are valid and ``bool`` has an underlying value of 0 or 1).
+
+   .. rationale::
+      :id: rat_kjFRrhpS8Wu6
+      :status: draft
+
+      Rust’s memory model requires that all bytes are initialized before a typed value may be constructed.
+      Reading uninitialized memory:
+
+      - creates undefined behavior for most types,
```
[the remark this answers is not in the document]
rcseacord: resolved

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,263 @@
...
+
+.. guideline:: Do not create values from uninitialized memory
+   :id: gui_uyp3mCj77FS8
+   :category: required
+   :status: draft
+   :release: <TODO>
+   :fls: fls_6lg0oaaopc26
+   :decidability: undecidable
+   :scope: system
+   :tags: undefined-behavior, unsafe
```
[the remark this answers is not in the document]
rcseacord: @felix91gr I'm not sure I understand what a defect type guideline is.  Does it mean any violation of this rule is definitely a defect in the code?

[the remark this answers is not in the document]
rcseacord: I understand the concepts.  Connecting terms to concepts can sometimes be challenging.  ;)

src/coding-guidelines/values.rst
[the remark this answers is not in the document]
rcseacord: I'll have to think on this a bit, as splitting guidelines is a lot of work.

src/coding-guidelines/values.rst
rcseacord: resolved

src/coding-guidelines/values.rst
```diff
@@ -5,3 +5,185 @@
...
+
+      Rust’s memory model treats all types except unions as having an invariant that all bytes must be initialized before a value may be constructed.
+      Reading uninitialized memory:
+
+      - creates undefined behavior for most types,
+      - may violate niche or discriminant validity,
+      - may create invalid pointer values, or
+      - may produce values that violate type invariants.
+      
+      The sole exception is that unions work like C unions: any union field may be read, even if it was never written.
```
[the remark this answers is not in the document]
rcseacord: resolved

src/coding-guidelines/values.rst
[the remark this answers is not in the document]
felix91gr: > I have changed the title to include reading because my understanding is that reading and calling assume_init are basically the same here.

That's the thing: they are not. But I don't think that matters much anyways.

# Why they are not the same

`MaybeUninit<T>` is very special for the compiler. `assume_init` is, among other things, a statement of "from here on, assume this memory is initialized". The compiler uses this fact to deduce things about the program, hence why it's UB to violate those assumptions. That memory never needs to be read for this to happen: it happens because a contract with the compiler has been broken.

Oh the other hand, reading from a `union` field is either well-defined or UB, depending on if the field is properly initialized or not.

They are not the same, except for the fact that both can cause UB if used poorly (which is a property of everything that inhabits in `unsafe`).

# Why I don't think it matters much anyway

`MaybeUninit<T>` is the main (almost only, in fact) avenue through which a programmer can create initialized, valid values of a type T, from uninitialized memory. It's a carefully defined API, and it has a tight integration with the compiler.

On the other hand, unions are a low-level primitive that can be used, among other things, to emulate `MaybeUninit<T>`. That's, in principle, how `MaybeUninit<T>` was constructed - you can see it's itself a `union`.

But unions are rather obscure in Unsafe Rust, and they have their own niche use cases. The contexts in which they are used are so different from `MaybeUninit<T>` that bundling them together seems like a mistake to me.

The process for safely working with them is more crude than how it is for `MaybeUninit<T>`, which I would also consider part of their mismatch if they were to be ruled about in the same guideline.

**Remember:** we could have a guideline called "Do not invoke UB", which would be more or less well-defined (we could say "MIRI under Tree Borrows", and that would be well-defined enough for our purposes), but it would be a terrible guideline because **so many different things** can do UB.

# What I'd do

Yeah, I'd separate it into two.

## One for `MaybeUninit<T>`

To make it more actionable, you could go straight ahead and make a guideline about how to use this type in a safe manner.

Basically, ruling on how to call the `assume_` APIs of `MaybeUninit<T>`:

* `assume_init`
* `assume_init_drop`
* `assume_init_mut`
* `assume_init_read`
* `assume_init_ref`

As well as:

* `array_assume_init`

Their documentation explains their safety invariants. The guideline would basically point out how and why to satisfy them.

## One for unions

Again, unions have their own niches and contexts. They are also WAY cruder to use than `MaybeUninit<T>`. I think the examples will show just how different they are.

I think unions merit their own rule. They are, after all, one of the few low-level primitives the language has.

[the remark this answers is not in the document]
felix91gr: PS: I retract from what I said about `MaybeUninit<T>` needing to be an exception to the `union` rule. Now that I'm looking at it: it may be an `union`, but its fields are all private. They cannot be read. Hence, it always satisfies the `union` rule :)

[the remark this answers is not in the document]
felix91gr: Ah, you're right. Okay, I'm not up to properly reviewing this then. @inkreasing or @workingjubilee can I interest you guys in reviewing this guideline?

I can help with the more form-related bits, like where should explanations go and where should the actionable items go. But I'll need the help of one of you, who are more experienced with this, to give the semantics a thumbs up.

Lemme know and I'll assign you as reviewer :3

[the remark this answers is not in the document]
workingjubilee: > I have changed the title to include reading because my understanding is that reading and calling `assume_init` are basically the same here.

This understanding is functionally correct: the important mechanical element is sometimes referred to as "transmuting" or "read-at-type". It is the low-level statement in the Rust abstract machine that moves some bytes and thinks of them as being bytes of a specific type as it does so. Bytes "at rest" have no such identity in the Rust AM, but they acquire a type while in motion because that often has an implication on the specific bytes that actually have to move. For instance, while it is definitely valid to copy all the bytes blindly, it may also be valid to e.g. move individual fields of a struct, instead of moving the entire struct-with-padding, which may be important when considering function calls in ABIs which do precisely that.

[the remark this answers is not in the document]
workingjubilee: The missing detail in your reading, @felix91gr, is that the `assume_init` documentation references the [initialization invariant](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#initialization-invariant) documentation on the type. That documentation describes that individual types have their own initialization invariant. We usually say something like "every byte has to be in the set of valid bitpatterns for that type". This is easily logically handled by tracking certain bytes as having "uninit" as a unique state that doesn't correspond to 0x00 through 0xFF. So for a type that allows uninit for a given byte, the typed read permits a read of the uninit bytes, as "0xUU" is in the valid set of bitpatterns. This does get slightly confusing because we don't consider it a bitpattern but we kinda... do...?

Anyway, it _is_ easy to read as being truly independent and context-free. We should probably change that.

src/coding-guidelines/values/gui_uyp3mCj77FS8.rst.inc
```diff
@@ -5,3 +5,263 @@
...
+   .. rationale::
+      :id: rat_kjFRrhpS8Wu6
+      :status: draft
+
+      Rust’s memory model requires that all bytes are initialized before a typed value may be constructed.
+      Reading uninitialized memory:
+
+      - creates undefined behavior for most types,
+      - may violate niche or discriminant validity,
+      - may create invalid pointer values, or
```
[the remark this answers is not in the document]
rcseacord: I simplified the text here.  The examples expand on pointer and reference validity.

src/coding-guidelines/values.rst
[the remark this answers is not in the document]
rcseacord: OK, I've resolved all of this in the new version

src/coding-guidelines/values/gui_uyp3mCj77FS8.rst.inc
```diff
@@ -0,0 +1,179 @@
...
+   Memory can remain uninitialized if it is not read as a type.
+   
+   Reading from a union is covered by a separate rule, `Do not read from union fields that may contain uninitialized bytes
+   <https://coding-guidelines.arewesafetycriticalyet.org/coding-guidelines/types-and-traits/index.html#gui_UnionPartialInit>`_.
+
+   Calling `assume_init <https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#method.assume_init>`_,
+   or related functions, is treated in the same manner as a typed read.
+   Calling these function when on memory that is not yet fully initialized causes immediate undefined behavior.
+   The memory must be properly initialized according to the requirements of the variable’s type.
+   For example, a variable of reference type must be aligned and non-null.
```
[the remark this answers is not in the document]
rcseacord: add this requirement and created a new noncompliant example

rcseacord (approved):
lgtm

src/coding-guidelines/values/gui_uyp3mCj77FS8.rst.inc
```diff
@@ -0,0 +1,278 @@
...
+         fn main() {
+             // Reading an invalid reference is undefined behavior
+             let _r: &u32 = unsafe { MaybeUninit::uninit().assume_init() }; // noncompliant
+         }
+
+   .. non_compliant_example::
+      :id: non_compl_ex_Qb5GqYTP6db3
+      :status: draft
+
+      This noncompliant example creates a reference from uninitialized memory.
```
inkreasing:
```suggestion
      This noncompliant example creates a reference without provenance.
```
Or maybe something that the reference is not dereferencable. (because the memory it points to doesn't exist)

src/coding-guidelines/values/gui_uyp3mCj77FS8.rst.inc
```diff
@@ -0,0 +1,278 @@
...
+      This compliant example defines a C-layout ``struct`` with:
+
+      * ``a``: 1 byte at offset 0
+      * 3 bytes of padding (to align ``b`` to 4 bytes)
+      * ``b``: 4 bytes at offset 4
+      * Total size: 8 bytes
+
+      The variable ``buf`` is a fully, zero-initialized 8-byte buffer.
+
+      The first wo bytes of ``buf`` are overwritten.
```
inkreasing:
```suggestion
      The first two bytes of ``buf`` are overwritten.
```

src/coding-guidelines/values/gui_uyp3mCj77FS8.rst.inc
```diff
@@ -0,0 +1,274 @@
...
+   <https://coding-guidelines.arewesafetycriticalyet.org/coding-guidelines/types-and-traits/index.html#gui_UnionPartialInit>`_.
+
+   Calling :std:`std::mem::MaybeUninit::assume_init`
+   :cite:`gui_uyp3mCj77FS8:MAYBEUNINIT-DOC` or any of the following related functions is treated in the same manner as a typed read:
+
+   * ``assume_init_drop``
+   * ``assume_init_mut``
+   * ``assume_init_read``
+   * ``assume_init_ref``
+   * ``array_assume_init``
```
felix91gr: Perhaps these APIs should also get linked here, much like `assume_init` is above

src/coding-guidelines/values/gui_uyp3mCj77FS8.rst.inc
```diff
@@ -0,0 +1,274 @@
...
+             x.write(42);
+             // SAFETY: 'x' is fully initialized
+             let _val = unsafe { x.assume_init() }; // compliant
+         }
+ 
+   .. non_compliant_example::
+      :id: non_compl_ex_Qb5GqYTP6db4
+      :status: draft
+
+      This noncompliant example creates a pointer from uninitialized memory.
```
felix91gr:
```suggestion
      This noncompliant example creates a pointer from uninitialized memory.
      
```

I'm adding a newline here because it feels to me like it's really a separate paragraph from the rest.

src/coding-guidelines/values/gui_uyp3mCj77FS8.rst.inc
```diff
@@ -0,0 +1,274 @@
...
+         fn main() {
+             // Undefined behavior creating a pointer from uninitialized memory
+             let _p: *const u32 = unsafe { MaybeUninit::uninit().assume_init() }; // noncompliant
+         }
+
+   .. non_compliant_example::
+      :id: non_compl_ex_Qb5GqYTP6db2
+      :status: draft
+
+      This noncompliant example creates a reference from uninitialized memory.
```
felix91gr:
```suggestion
      This noncompliant example creates a reference from uninitialized memory.
      
```

Paragraph :)

src/coding-guidelines/values/gui_uyp3mCj77FS8.rst.inc
```diff
@@ -0,0 +1,274 @@
...
+
+   .. compliant_example::
+      :id: compl_ex_Ke869nSXuShV
+      :status: draft
+
+      This compliant example creates an uninitialized variable ``x`` of type ``MaybeUninit<u64>``.
+      The code calls the ``write`` function to write the value 42 to ``x``.
+      The call to ``assume_init`` asserts that the value is initialized and extracts the value of type ``u64``.
+      This call is compliant with this rule because the memory has been properly initialized by the call to ``write(42)``.
+      This is the canonical safe pattern for using ``MaybeUninit``.
```
felix91gr:
```suggestion

      This call to ``assume_init`` is compliant with the rule because the memory of `x` has been properly initialized by the call to ``write(42)``.
      
      This is the canonically safe pattern for using ``MaybeUninit``.
```

felix91gr (changes requested):
Sorry I took so long. I hope this helps.

I think the examples are pretty good in general; I'm mostly suggesting improvements to the presentation plus a couple of details here and there.

src/coding-guidelines/values/gui_uyp3mCj77FS8.rst.inc
rcseacord: added some links, now we'll see if it builds.

rcseacord (approved):
lgtm

src/coding-guidelines/values/gui_uyp3mCj77FS8.rst.inc
```diff
@@ -0,0 +1,274 @@
...
+             // Undefined behavior creating a pointer from uninitialized memory
+             let _p: *const u32 = unsafe { MaybeUninit::uninit().assume_init() }; // noncompliant
+         }
+
+   .. non_compliant_example::
+      :id: non_compl_ex_Qb5GqYTP6db2
+      :status: draft
+
+      This noncompliant example creates a reference from uninitialized memory.
+      Creating a reference from arbitrary or uninitialized bytes is undefined behavior :cite:`gui_uyp3mCj77FS8:RUST-REF-BEHAVIOR`.
```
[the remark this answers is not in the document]
felix91gr: @rcseacord I'm still seeing this. Can you see it?

felix91gr (commented):
I've updated my comments; if I remember this correctly, I think this PR is pretty close to done.

Only a couple ends left to tie.

PR #244, opened 2025-12-05
Update shifting operations guidelines in expressions.rst

rcseacord:
Clarify rules for shifting operations in Rust, emphasizing the importance of using checked functions and addressing inconsistent behavior in different compilation modes.

### DIFF
--- a/src/coding-guidelines/expressions.rst
+++ b/src/coding-guidelines/expressions.rst
@@ -780,7 +780,7 @@ Expressions
           /* ... */
         }
 
-.. guideline:: Avoid out-or-range shifts
+.. guideline:: Do not shift an expression by a negative number of bits or by greater than or equal to the bitwidth of the operand
     :id: gui_RHvQj8BHlz9b 
     :category: advisory
     :status: draft
@@ -991,7 +991,7 @@ Expressions
                  }
              }
 
-.. guideline:: Do not shift an expression by a negative number of bits or by greater than or equal to the number of bits that exist in the operand
+.. guideline:: Avoid out-or-range shifts
     :id: gui_LvmzGKdsAgI5 
     :category: mandatory
     :status: draft


src/coding-guidelines/expressions.rst
```diff
@@ -815,37 +814,36 @@ Expressions
...
-          | Release          | Shifts normally | Shifts by ``M mod N`` | Shifts by ``M mod N`` |
+          | Release          | Shifts ``M`` positions | Shifts by ``M mod N`` | Shifts by ``M mod N`` |
           +------------------+-----------------+-----------------------+-----------------------+
 
           ..
 
-             Note: the behavior is exactly the same for the ``>>`` operator.
-
-
           Panicking in ``Debug`` is an issue by itself, however, a perhaps larger issue there is that its behavior is different from that of ``Release``. Such inconsistencies aren't acceptable in Safety Critical scenarios.
```
rcseacord: as you point out, the UB here while generally innocuous  never makes sense.  consequently, if it happens in debug I probably want to know about it because it's a defect that requires repair.  this is different from arithmetic overflow where maybe the developer wants wrap around behavior but now can't test the code in debug because it panics. 

For these reasons, I don't think the inconsistency between debug and release is beneficial and not actually a problem (in this case).

rcseacord: @felix91gr I think this entire section on release versus debug behavior should go.  Let me know what you think.

src/coding-guidelines/expressions.rst
rcseacord: Don't ever suggest there is only one solution.  What about shifting by a ranged integer type, for example?

rcseacord: only shifting by unsigned integers is a good idea and maybe should be part of some rule

src/coding-guidelines/expressions.rst
```diff
@@ -783,15 +782,17 @@ Expressions
...
     :release: 1.7.0-latest
     :fls: fls_sru4wi5jomoe
     :decidability: decidable
     :scope: module
     :tags: numerics, reduce-human-error, maintainability, portability, surprising-behavior, subset
 
-    In particular, the user should only perform left shifts via the `checked_shl <https://doc.rust-lang.org/core/index.html?search=%22checked_shl%22>`_ function and right shifts via the `checked_shr <https://doc.rust-lang.org/core/index.html?search=%22checked_shr%22>`_ function. Both of these functions exist in `core <https://doc.rust-lang.org/core/index.html>`_.
+    In particular, the user should only perform left shifts via the `checked_shl <https://doc.rust-lang.org/core/index.html?search=%22checked_shl%22>`_
+    function and right shifts via the `checked_shr <https://doc.rust-lang.org/core/index.html?search=%22checked_shr%22>`_ function.
+    Both of these functions exist in `core <https://doc.rust-lang.org/core/index.html>`_.
```
felix91gr: @rcseacord you added these newlines to account for this bug, right? https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/issues/265

src/coding-guidelines/expressions.rst
felix91gr: It's part of the Defect rule, actually!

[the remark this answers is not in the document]
felix91gr: > Not getting a result isn't always an option.

We talked about this today. Did we reach a consensus here?

I stand by how for the division and modulo operation, if one is to use a divisor `x` in a function `f(x)` and has the possibility of inputting `x = 0`, then one can make `f` give out a specific value for `f(0)` instead of panicking.

It is the same for functions `g` that use a shifting amount `y` for types `uN`. The user always has the option to divide the domain of `g` and map `y < 0` and `y >= N` in `g(y)` to something other than panics.

> It's basically the equivalent of saturation semantics which we allow as a compliant solution.

In what way is it like saturation semantics? Saturation would either mean that:
* `x << y ----> x` for `y < 0` and `x << y ----> x << N - 1` for `y >= N`, or that
* `x << y ----> x` for `y < 0` and `x << y ----> uN::MAX` for `y >= N`

This isn't like wrapping semantics, either. It's... something very weird instead. I mean, apart from being inconsistent, it's just very weird.

It makes complete sense for rotating shifts, since there a higher `y` just means doing 1 or more full rotations, which are trivial and thus can be eliminated by doing `y = y % N`. But for normal shifts... it's just nonsense.

If anything, I would actually expect `x << y` to give me `0`, since the bits of `x` are completely replaced by zeros:

```
// Example: x : u8 = abcdefgh
// x sits in an infinite ocean of 0s.
...0000000000abcdefgh000000...
             ^      ^---------- x ends here
             |      
             -- x begins here          

// We compute x << 8,
// moving the bits of x 8 spaces to the left
...00abcdefgh00000000000000...
             ^      ^---------- x ends here
             |      
             -- x begins here          
// x is now 0.
```

Which would be consistent with how `<<` and `>>` work otherwise. It is **extremely** weird that eg `1u8 << 6` is `64`, `1u8 << 7` is `128`, and yet `1u8 << 8` is somehow `1`. It doesn't even follow multiplication wrapping, which would make `1u8 << 8` equal to `256`, and therefore `0`. You know what I mean?

src/coding-guidelines/expressions.rst
```diff
@@ -781,21 +780,42 @@ Expressions
...
+    the shift right expression ``lhs >> rhs`` evaluates to the value of the left operand ``lhs`` 
+    whose bits are shifted right by the number of positions speicifed by the right operand ``rhs``.
+    If the type of the left operand is any signed integer type and is negative,
+    the vacated bits are filled with ones.
+    Otherwise, vacated bits are filled with zeros.
+    The expression ``lhs >> rhs`` evaluates to :math:`\mathrm{lhs} / 2^{\mathrm{rhs}}`,
+    cast to the type of the left operand.
+    If the value of the right operand is negative,
+    greater than or equal to the width of the left operand,
+    then the operation results in an arithmetic overflow.
```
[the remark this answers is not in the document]
rcseacord: I was also debating the "integer type" line so I'll remove it.  Only integer types are possible.

src/coding-guidelines/expressions.rst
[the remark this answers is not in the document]
rcseacord: yeah, it's not saturation as we discussed.  it's just an example of an approach where instead of having an error you try to produce an in-range value.

felix91gr:
@rcseacord I found something you can look at while I'm offline.

The page build script is giving errors on these two lines:

```
6:03:42 AM: /opt/build/repo/src/coding-guidelines/expressions.rst:783: WARNING: Need could not be created: Status None not in 'needs_statuses'. [needs.create_need]
6:03:42 AM: /opt/build/repo/src/coding-guidelines/expressions.rst:994: WARNING: Need could not be created: Status None not in 'needs_statuses'. [needs.create_need]
```

I don't know what they mean, but if you can manage to build the site locally, then these errors should disappear :)

PR #256, opened 2025-12-06
[Coding Guideline]: Do Not Depend on Function Pointer Identity

rcseacord:
Coding guidelines based on https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/issues/255

### DIFF
--- a/src/coding-guidelines/types-and-traits/gui_QbvIknd9qNF6.rst
+++ b/src/coding-guidelines/types-and-traits/gui_QbvIknd9qNF6.rst
@@ -360,12 +360,11 @@ Do not directly or indirectly compare function pointers
 
           fn get_a() -> (u8, u8, u8, u8) {
               let mut a = MyMaybeUninit { init: (0, 0, 0, 0) };
-              let addr1 = write_first as usize;
-              let addr2 = write_second as usize;
+              let addr1 = write_first as *const ();
+              let addr2 = write_second as *const ();
               if addr1 == addr2 {
                   unsafe {
-                      let ptr = addr1 as *const ();
-                      let f: fn(&mut MyMaybeUninit) = core::mem::transmute(ptr);
+                      let f: fn(&mut MyMaybeUninit) = core::mem::transmute(addr1);
                       f(&mut a);
                   }
               }


src/coding-guidelines/types-and-traits.rst
```diff
@@ -63,3 +64,225 @@ Types and Traits
...
+    :category: required
+    :status: draft
+    :release: unclear-latest
+    :fls: fls_1kg1mknf4yx7
+    :decidability: decidable
+    :scope: system
+    :tags: surprising-behavior
+
+    Do not rely on the equality or stable identity of function pointers originating from different crates or that may be inlined,
+    duplicated, or instantiated differently across compilation units, codegen units, or optimization profiles.
```
saethlin: I'm not sure why you're mentioning optimization profiles here; I don't think it needs to be pointed out that functions may change if you recompile with different compiler settings.

A compilation unit and codegen unit is the same thing for these purposes, and I at least would prefer that people say "codegen unit" because that's what the option is called in the rustc CLI and the Cargo profiles.

src/coding-guidelines/types-and-traits.rst
```diff
@@ -63,3 +64,225 @@ Types and Traits
...
+    Avoid assumptions about low-level metadata (such as symbol addresses) unless explicitly guaranteed by the Ferrocene Language Specification (FLS).
+    Function address identity is not guaranteed by Rust and must not be treated as stable.
+    Rust’s ``fn`` type is a zero-sized function item promoted to a function pointer, whose address is determined by the compiler backend.
+    When a function resides in a different crate, or when optimizations such as inlining,
+    link-time optimization, or codegen-unit partitioning are enabled,
+    the compiler may generate multiple distinct code instances for the same function or alter the address at which it is emitted.
+
+    Consequently, the following operations are not reliable:
+
+    - Comparing function pointers for equality (``fn1 == fn2``)
```
saethlin: This is reliable for functions which are `#[no_mangle]`

src/coding-guidelines/types-and-traits.rst
```diff
@@ -63,3 +64,225 @@ Types and Traits
...
+    - Assuming a unique function address
+    - Using function pointers as identity keys (e.g., in maps, registries, matchers)
+    - Matching behavior based on function address
```
saethlin: If you tell the linker to put a (`#[no_mangle]`) function at a specific address, I think it is reasonable to rely on that in your Rust code.

src/coding-guidelines/types-and-traits.rst
```diff
@@ -63,3 +64,225 @@ Types and Traits
...
+        :id: rat_xcVE5Hfnbb2u 
+        :status: draft
+
+        Compiler optimizations may cause function pointers originating from different crates to lose stable identity.
+        Observed behaviors include:
+
+        - Cross-crate inlining producing multiple code instantiations
+        - Codegen-unit separation causing function emission in multiple units
+        - Incremental builds producing variant symbol addresses
+        - Link-time optimization merging or splitting functions unpredictably
```
saethlin: Nit: functions can be merged even when LTO is not enabled

src/coding-guidelines/types-and-traits.rst
```diff
@@ -63,3 +64,225 @@ Types and Traits
...
+
+        This behavior has resulted in real-world issues,
+        such as the bug reported in `rust-lang/rust#117047 <https://github.com/rust-lang/rust/issues/117047>`_,
+        where function pointer comparisons unexpectedly failed due to cross-crate inlining.
```
saethlin: Note that in the phrase "cross crate inlining", the term "inline" does not refer to the optimization where a call is replaced by the body of the callee, it refers to a strategy of lowering compiler IR in the caller crate instead of linking to it in the callee.

The term "inline" is used very confusingly. So in this case I would just say

>         where function pointer comparisons unexpectedly failed because the function in question was instantiated multiple times.

saethlin: Perhaps "instantiated" is a confusing term to those with a primarily C background, but it's supposed to be intuitive to those with more of a C++ mindset 🙃

src/coding-guidelines/types-and-traits.rst
```diff
@@ -63,3 +64,225 @@ Types and Traits
...
+    Rust’s ``fn`` type is a zero-sized function item promoted to a function pointer, whose address is determined by the compiler backend.
+    When a function resides in a different crate, or when optimizations such as inlining,
+    link-time optimization, or codegen-unit partitioning are enabled,
+    the compiler may generate multiple distinct code instances for the same function or alter the address at which it is emitted.
+
+    Consequently, the following operations are not reliable:
+
+    - Comparing function pointers for equality (``fn1 == fn2``)
+    - Assuming a unique function address
+    - Using function pointers as identity keys (e.g., in maps, registries, matchers)
```
programmerjake: you can still use function pointers as keys, they just don't have a one-to-one relationship to function items, so you can't rely on a function item giving the same function pointer or different items giving different pointers. (ignoring `#[no_mangle]`). so e.g. you can have a `HashSet<fn()>` and it works just fine as long as you don't assume inserting any particular function item does anything other than guaranteeing calling one or more of the entries is equivalent to calling that function item.

src/coding-guidelines/types-and-traits.rst
rcseacord: resolved

src/coding-guidelines/types-and-traits.rst
rcseacord: @programmerjake i think this has been refuted over in [#general > UB in Safe Rust @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/122651-general/topic/UB.20in.20Safe.20Rust/near/562267154)

src/coding-guidelines/types-and-traits.rst
rcseacord: resolved

src/coding-guidelines/types-and-traits.rst
rcseacord: I can grok instantiated just fine.  Anyway, the target audience is rust programmer.
Resolved.

felix91gr:
@rcseacord you'll want to consider this discussion before the guideline is ready for merge https://github.com/rust-lang/unsafe-code-guidelines/issues/589, since it addresses the same topic

src/coding-guidelines/types-and-traits/gui_QbvIknd9qNF6.rst.inc
```diff
@@ -0,0 +1,257 @@
...
+         }
+
+         #[unsafe(no_mangle)]
+         fn bar(x: *mut i32, y: *mut i32) {
+             unsafe {
+                 x.write(y.read());
+             }
+         }
+
+         For specific backends, That is, the two functions are deduplicated, and therefore have the same address.
```
rcseacord:
```suggestion
         These functions are deduplicated for specific backends and have the same address.
```

plaindocs:
Hey @vccjgust would you like to claim this review, or let someone else grab it?

vccjgust:
Hi @plaindocs, sorry didn't know it needed claiming since it was already assigned to me (or so I thought). I'm taking a look at it as we speak! 😃

PR #270, opened 2025-12-11
[Coding Guideline]:  Ensure reads of union fields produce valid values

rcseacord:
added new guideline on unions


manhatsu:
I moved the commits here to a new feature branch rcseacord:doc/union-field-varidation. The corresponding PR is #300.

PR #305, opened 2025-12-17
[Coding Guideline]: Ensure all loops have a termination condition that is provably reachable

rcseacord:


rcseacord:
> Interesting rule. Are we sure there aren't safety-critical systems which have at least one thread that loops forever? @rcseacord

@felix91gr quite possibly.  the philosophy behind this rule (which is quite common in MISRA) is that this is pretty dangerous and so it should always be flagged by the analyzer.  any time you have an infinite loop in your code will require a deviation permit, meaning that that infinite loops will always be scrutinized more carefully (presumably by a senior engineer who audits these things).  

Presumably there aren't many of these in any given system (I would guess at most one) so the burden isn't that high.

PLeVasseur:
@guidelines-bot claim

PLeVasseur:
@guidelines-bot label +chapter-expressions

PLeVasseur:
@guidelines-bot label +chapter: expressions +decidability: decidable

PLeVasseur:
@guidelines-bot label +chapter: values

PLeVasseur:
@guidelines-bot pass! testing

PLeVasseur:
@guideilnes-bot /release

PLeVasseur:
@guidelines-bot /status

PR #331, opened 2025-12-18
docs: update CONTRIBUTING.md with guidelines-bot details and review expectations

PLeVasseur:


CONTRIBUTING.md
```diff
@@ -95,6 +100,8 @@ contents converted to reStructuredText.
 
 Within 14 days of your submission, a member of the Coding Guidelines Subcommittee should give you a first review. You'll work with them (and other members) to flesh out the concept and ensure the guideline is well prepared for a Pull Request.
 
+> **Note:** A reviewer is automatically assigned from the pool of Producers using a round-robin system. See [Reviewer Bot Commands](#reviewer-bot-commands) for details.
```
felix91gr: We might want to clarify what we mean by a round-robin system.

I think I know what we mean here... something like this, perhaps?

1. The unassigned producers are put in a queue,
2. When a review is complete, its producer (reviewer) is put back at the end of the queue,
3. And all of this results in what is effectively a complete and unbiased rotation of the producer group until any given producer is assigned to review once again.

Did I get it right?

If so, what happens if the queue is empty (because every producer is either on break or reviewing) and a new guideline proposal is opened?

felix91gr:
@guidelines-bot /r? @AlexCeleste

CONTRIBUTING.md
felix91gr: Hell yeah

CONTRIBUTING.md
```diff
@@ -169,6 +186,198 @@ Once the coding guideline contents have passed review, a subcommittee member wil
...
+- Skim the relevant FLS section and any linked documentation
+- Read through the guideline with fresh eyes, noting what *does* make sense
+- Search for a blog post or example that illuminates the concept
+- Try compiling and tweaking the code examples yourself
+
+You may find that an hour of targeted exploration is enough to provide meaningful feedback, even if you're not an expert. Catching unclear explanations, spotting typos, or asking "what does this term mean?"—these contributions are valuable precisely *because* you're approaching the material without deep familiarity.
+
+That said, `/pass` exists for good reason. If after an honest effort the material remains opaque, or if the guideline requires genuine expertise you don't have (and can't reasonably acquire in an hour), passing to someone better suited is the right call. The goal is thoughtful engagement, not struggling through a review you can't meaningfully contribute to.
+
+#### Examples of Valid Reasons to Pass
```
felix91gr:
```suggestion
### Examples of Valid Reasons to Pass
```

(S.A. stands for See Above)

felix91gr (changes requested):
Nice work with the remaining tidbits :)

I've added suggestions to accommodate for the changes in headings. Two items that were of level 3 now should be bumped up to level 2 (since they're not commands themselves), and the headings under those should be bumped up to level 3 (the entire subtrees go up 1 level).

felix91gr:
@PLeVasseur we should also add the `/rectify` command to these docs

felix91gr:
@guidelines-bot /unassign

PR #353, opened 2026-01-01
TEST DO NOT MERGE - fix: make it so that merge queue will fail even if a dependency failed

PLeVasseur:


PLeVasseur:
problem confirmed

PR #354, opened 2026-01-01
TEST DO NOT MERGE - Check with ci fix

PLeVasseur:


PLeVasseur:
Confirms that indeed, with no change other than fixing CI this will fail now

PR #368, opened 2026-01-10
[Coding Guideline] Revise guideline on unsafe code in macros

rcseacord:
Updated guideline on unsafe code in macros to clarify the importance of preserving 'unsafe' token visibility. Added examples of compliant and non-compliant macro implementations.

### DIFF
--- a/src/coding-guidelines/macros/gui_FRLaMIMb4t3S.rst
+++ b/src/coding-guidelines/macros/gui_FRLaMIMb4t3S.rst
@@ -3,53 +3,72 @@
 
 .. default-domain:: coding-guidelines
 
-Do not hide unsafe blocks within macro expansions
-=================================================
+Do not hide unsafe code in macros
+================================
 
-.. guideline:: Do not hide unsafe blocks within macro expansions
+.. guideline:: Do not hide unsafe code in macros
    :id: gui_FRLaMIMb4t3S
-   :category: required
+   :category: advisory
    :status: draft
-   :release: todo
+   :release: unclear-latest
    :fls: fls_4vjbkm4ceymk
-   :decidability: todo
-   :scope: todo
-   :tags: reduce-human-error
+   :decidability: decidable
+   :scope: module
+   :tags: safety, reduce-human-error
 
-   Description of the guideline goes here.
+   Do not hide unsafe code in macro definitions.
+   Macros that expand to unsafe code should preserve the ``unsafe`` token visibility.
 
    .. rationale::
-      :id: rat_WJubG7KuUDLW
+      :id: rat_s7ffMlPUFt77
       :status: draft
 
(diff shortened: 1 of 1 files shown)


felix91gr:
I opened a Zulip thread to ask about this: _[How should one combine `unsafe` and declarative macros?](https://rust-lang.zulipchat.com/#narrow/channel/122651-general/topic/How.20should.20one.20combine.20.60unsafe.60.20and.20declarative.20macros.3F/with/568625091)_.

I wanted to know how useful this pattern (of having an `unsafe` block inside of a macro's code) is.

And it seems like it is quite useful, and necessary in some cases:

A very simple example by scottmcm:

```rust
let slice = $crate::MyUnsafeTrait::get_me_the_thing($e);
// SAFETY: `MyUnsafeTrait` promises that `get_me_the_thing` returns a non-empty slice
unsafe { slice.get_unchecked(0) }
```

Another example, this time from `std` (explained by y21):

> > Are there things that this unlocks, that are otherwise impossible or unreasonable to achieve with e.g. invoking the macro inside of an unsafe block instead?
> 
> I think the [`pin!`](https://doc.rust-lang.org/stable/std/pin/macro.pin.html) macro is also a good example. It calls the unsafe function `Pin::new_unchecked` in an `unsafe` block, and it can do so soundly because it takes ownership of the value and only gives back a `Pin<&mut _>`, ensuring that it stays pinned forever. This only works because it's a macro, and there's no reason to require an `unsafe` block at the use site because the safety contract of `Pin::new_unchecked` is upheld by the macro itself; there's nothing that a user needs to uphold when invoking this macro. So an unsafe block inside of the macro is perfectly valid there

mdabir1203:
Yes I will

mdabir1203 (changes requested):
One more thing I am confused is this :  FLS paragraph ID ??

src/coding-guidelines/macros/gui_FRLaMIMb4t3S.rst
```diff
@@ -3,53 +3,72 @@
...
-      Explanation of why this guideline is important.
+      Macros that generate unsafe code obscure safety-critical code, making the code more difficult to review and audit.
 
    .. non_compliant_example::
-      :id: non_compl_ex_AyFnP0lJLHxi
+      :id: non_compl_ex_YNX7AnWENTu7
       :status: draft
 
-      Explanation of code example.
+      Macros that generate unsafe code without preserving the ``unsafe`` token visibility obscure safety-critical code.
```
mdabir1203: Macros that generate unsafe code without preserving the ``unsafe`` token visibility obscure safety-critical code.

My suggestion would be to make the ending of the sentence to be more precise and clear :

 Macros that generate unsafe code without preserving the ``unsafe`` token visibility _for obscure safety-critical code._

PR #372, opened 2026-01-17
Update spec.lock, 3 guidelines related but not affected

felix91gr:
1) The two div-by-zero guidelines are related to https://rust-lang.github.io/fls/expressions.html#fls_Q9dhNiICGIfr
This element of the FLS hasn't significantly changed (dividing by 0 still results in panic).

2) One very old pointer guideline is related to https://rust-lang.github.io/fls/expressions.html#fls_9wgldua1u8yt
This element, as far as I can tell, hasn't changed significantly either (dereferencing dangling or unaligned pointers has been UB for a while).

### DIFF
--- a/src/spec.lock
+++ b/src/spec.lock
@@ -1,122 +1,227 @@
 {
     "documents": [
         {
+            "informational": true,
+            "link": "changelog.html",
+            "sections": [],
+            "title": "Changelog"
+        },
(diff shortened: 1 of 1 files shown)


PLeVasseur (approved):
Thanks for picking this up, @felix91gr!

PR #373, opened 2026-01-17
Fix code example qualifiers

felix91gr:
Some code examples weren't marked with the right kind of behavior they should've had.

The code that handled Playground responses to panicking code wasn't behaving correctly.

My guess is that the checking for `result.success` should've been inverted (I believe that `!success` means something failed in the code, which might've been a compilation or a run-time failure. If that's correct, then I think my change is the right fix. CC @PLeVasseur lemme know what you think of that one).

This PR depends on https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/pull/372 's first commit. If #372 is merged first, then I believe this one should need no changes to accommodate for the FLS updating.

### DIFF
--- a/src/coding-guidelines/expressions/gui_RHvQj8BHlz9b.rst.inc
+++ b/src/coding-guidelines/expressions/gui_RHvQj8BHlz9b.rst.inc
@@ -88,6 +88,7 @@
         This noncompliant example shifts by a negative value (-1) and also by greater than or equal to the number of bits that exist in the left operand (40):.
 
         .. rust-example::
+            :should_panic:
 
             fn main() {
                 let bits : u32 = 61;
@@ -102,7 +103,7 @@
         :id: non_compl_ex_mvkgTL3kulZ5 
         :status: draft
 
-        This noncompliant example test the value of ``sh`` to ensure the value of the right operand is negative or greater 
+        This noncompliant example tests the value of ``sh`` to ensure the value of the right operand is negative or greater 
         than or equal to the width of the left operand.
 
         .. rust-example::


src/coding-guidelines/types-and-traits/gui_0cuTYG8RVYjg.rst.inc
```diff
@@ -77,7 +77,7 @@
       This noncompliant example reads an invalid Unicode value from a ``union`` field of type ``char`` .
 
       .. rust-example::
-         :miri:
+         :miri: expect_ub
```
PLeVasseur: True, this should be `expect_ub`.

src/coding-guidelines/expressions/gui_RHvQj8BHlz9b.rst.inc
```diff
@@ -88,6 +88,7 @@
         This noncompliant example shifts by a negative value (-1) and also by greater than or equal to the number of bits that exist in the left operand (40):.
 
         .. rust-example::
+            :should_panic:
```
PLeVasseur: Good catch, yes, this should be `should_panic`.

exts/coding_guidelines/rust_examples.py
```diff
@@ -1294,7 +1335,7 @@ def get_js_content() -> str:
                 }
                 break;
             case 'should_panic':
-                if (success && ((result.stderr && result.stderr.includes('panicked')) || 
+                if (!success && ((result.stderr && result.stderr.includes('panicked')) ||
```
PLeVasseur: Good observation. This is now correct.

PLeVasseur (approved):
Nice one finding the logic error and fixing the examples 🤝

Issue #314, opened 2025-12-17
📊 Reviewer Bot State (DO NOT CLOSE)


PLeVasseur:
## 📊 Reviewer Bot State

> ⚠️ **DO NOT EDIT MANUALLY** - This issue is automatically maintained by the reviewer bot.
> Use bot commands instead (see [[CONTRIBUTING.md](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/blob/main/CONTRIBUTING.md)](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/blob/main/CONTRIBUTING.md) for details).

This issue tracks the round-robin assignment of reviewers for coding guidelines.

### Current State

```yaml
last_updated: null
current_index: 0
queue: []
pass_until: []
recent_assignments: []
```

### What This Tracks

- **queue**: Active reviewers in rotation order
- **current_index**: Position in queue (who's next)
- **pass_until**: Reviewers temporarily away with return dates
- **recent_assignments**: Last 20 assignments for visibility

PR #405, opened 2026-02-10
fix(reviewer-bot): mutation-intent routing and git-ref CAS state lock

PLeVasseur:
## Summary
- Fixes the reviewer-bot CI regression caused by unsupported conditional writes (`If-Match`) on Issue `PATCH`.
- Replaces issue-body lease locking with a git-ref CAS lock so all mutating reviewer-bot paths are serialized safely.
- Switches lock gating from event-name based to mutation-intent based, so fork review events can defer cleanly while trusted reconcile remains lock-protected.

## Root Cause
- The prior hardening attempted lease acquire with `If-Match` against the state issue update endpoint.
- GitHub rejects conditional headers for this unsafe request path, which caused lock acquire to fail before event handling.
- For fork `pull_request_review` events, that failure happened before deferral/handoff logic, so reconcile never ran.

## What This PR Does Now
- Adds mutation-intent routing in `main()` with explicit intents:
  - `mutating`
  - `non_mutating_defer`
  - `non_mutating_readonly`
- Acquires lock only for `mutating` paths.
- Implements git-ref CAS locking in `scripts/reviewer_bot.py`:
  - lock ref bootstrap/read (`refs/heads/reviewer-bot-state-lock`)
  - lock/unlock commits with owner token and expiry metadata
  - fast-forward ref update (`force=false`) with retry/backoff on conflicts
  - lease freshness renewal before state persistence
- Removes lock-time dependence on issue `If-Match`; state saves now occur under lock without unsupported conditional headers.
- Keeps fail-closed behavior for mutating paths if lock or save fails.
- Updates `.github/workflows/reviewer-bot-reconcile.yml` to `contents: write` so reconcile can participate in lock acquire/release.
- Removes workflow-level per-item `concurrency` from `.github/workflows/reviewer-bot.yml` so serialization is lock-driven and does not rely on pending-run replacement.
- Retains existing triage escalation behavior from earlier commits in this PR.

## Fork PR Approval Flow (Current)
1. Fork `pull_request_review/submitted` run is non-mutating and defers.
2. First-hop run succeeds and publishes reconcile context.
3. Trusted `workflow_run` reconcile acquires CAS lock, mutates state issue `#314`, and releases lock.
4. `/rectify` remains available as manual fallback.

## Validation
- `uv run ruff check --fix scripts/reviewer_bot.py .github/reviewer-bot-tests/test_reviewer_bot.py`
- `uv run pytest .github/reviewer-bot-tests/test_reviewer_bot.py` (93 passed)
- PR #405 required CI checks are green.

### DIFF
--- a/.github/reviewer-bot-tests/test_reviewer_bot.py
+++ b/.github/reviewer-bot-tests/test_reviewer_bot.py
@@ -336,22 +336,26 @@ def test_acquire_state_issue_lease_lock_success(monkeypatch):
     monkeypatch.setenv("WORKFLOW_JOB_NAME", "reviewer-bot")
     monkeypatch.setattr(reviewer_bot.random, "uniform", lambda a, b: 0.0)
     monkeypatch.setattr(reviewer_bot.time, "sleep", lambda _: None)
-
-    body = reviewer_bot.render_state_issue_body(make_state(), reviewer_bot.clear_lock_metadata())
     monkeypatch.setattr(
         reviewer_bot,
-        "get_state_issue_snapshot",
-        lambda: reviewer_bot.StateIssueSnapshot(
-            body=body,
-            etag='"etag"',
-            html_url="https://example.com/issues/314",
+        "get_lock_ref_snapshot",
+        lambda: ("parent-sha", "tree-sha", reviewer_bot.clear_lock_metadata()),
+    )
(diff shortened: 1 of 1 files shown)


felix91gr (approved):
(see my last comment; I think this is the best course of action even in the worst case)

PR #539, opened 2026-03-22
Removed blanket forbiddance of macro usage

### LINKED ISSUE #367: [Old Guideline Errors and Merits] Blanket forbiddance of macro usage
This is part of a larger endeavor to fix numerous issues with our Macros chapter.
[Go to the index here](https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/issues/363)

There are three rules that I'm grouping into this one issue, because they feel like they kind of fail in the same way.

Please feel free to leave feedback on this!

The three rules are these:

* [Shall not invoke macros](https://coding-guidelines.arewesafetycriticalyet.org/coding-guidelines/macros/index.html#gui_a1mHfjgKk4Xr)
* [Shall not use Declarative Macros](https://coding-guidelines.arewesafetycriticalyet.org/coding-guidelines/macros/index.html#gui_h0uG1C9ZjryA)
* [Shall not write code that expands macros](https://coding-guidelines.arewesafetycriticalyet.org/coding-guidelines/macros/index.html#gui_uuDOArzyO3Qw)

This is the list of problems they have:
1. They are all empty / stub rules.
    I would love to see what the reasoning is behind them, because one never knows what insights one might find.

    But I digress. They are empty, which means they need to be filled before they can be useful.
2. "Shall not invoke macros" and "Shall not write code that expands macros" are the same thing. The former is the compile-time behavior of the latter.
3. The big one: **they are blanket prohibitions.**
    * Not invoking macros is pretty bad. Making functional Rust code without macros is extremely hard, and I wouldn't be surprised if the resulting code had more errors than the state-of-the art. There's so much code that good macros streamline, that it is kind of a super-power of Rust, for better or worse (mostly for the better).
    * Not using Declarative Macros... well, I mean, you gotta use one of the two kinds of macros, right? And Declarative Macros tend to be the default kind of macro to use. I don't think it's reasonable to forbid the default, unless this default is actively harmful.

So, anyway. I would just Retire these three guidelines. I don't think we can use them.

felix91gr:
Closes #367


felix91gr:
CC @PLeVasseur, I had these pending, finally got around to deleting them in a branch. Whole thing tool like 10 minutes x'D

plaindocs:
@felix91gr do you need the `uv.lock` in here? Cleaner not to right?

PR #1216, opened 2026-07-08
Fix: prevent make.py from updating uv.lock (#387)

### LINKED ISSUE #387: Fix make.py updating uv.lock
Currently when you run `./make.py` to generate the guidelines it modifies `uv.lock` instead of installing what is listed in the `uv.lock`, which results in a dirty git repo.

This might be related to the shebang in `make.py`:

```
#!/usr/bin/env -S uv run
``` 

Tagging @PLeVasseur

MarcosBorgesPhD:
Harmonize the way we run `make.py` to be cross-platform and prevent it from updating the `uv.lock` which results in a dirty git repo.

Replace the usage:

- macOS/Linux: `./make.py`
- Windows: `uv run make.py`

With:

- macOS/Linux/Windows: `uv run --frozen make.py`

Closes #387


MarcosBorgesPhD:
@plaindocs, since I removed the shebang `#!/usr/bin/env -S uv run` from `make.py`, the Netlify build command (`sh && ./make.py`) will now fail. 

Do you prefer to fix this on the Netlify side (by changing the build command), or should we keep the shebang in `make.py`?

PLeVasseur:
> Do you prefer to fix this on the Netlify side (by changing the build command), or should we keep the shebang in `make.py`?

I think we should fix this on the Netlify side, that feels more correct.

MarcosBorgesPhD:
> I think we should fix this on the Netlify side, that feels more correct.

@PLeVasseur, @plaindocs, who can help us with the Netlify fix?

It is necessary to replace the calls `./make.py` with `uv run --frozen make.py`.

Also, we need to update the `spec.lock`! How do we do it? Just send a PR with the updated version?

plaindocs:
Netlify is a @PLeVasseur thing afaik.

Re updating the spec lock, yes, there are [some docs](https://github.com/Safety-Critical-Rust-Consortium/safety-critical-rust-coding-guidelines#build-breaking-due-to-out-dated-spec-lock-file)

PLeVasseur:
@JoelMarcey -- could you help us out with the Netlify update? 🙏 

@MarcosBorgesPhD, @plaindocs -- could you give the instructions on what needs updating?

I see there's already some description up in this [comment](https://github.com/Safety-Critical-Rust-Consortium/safety-critical-rust-coding-guidelines/pull/1216#issuecomment-4983685568), but if @JoelMarcey has further questions, support here would be appreciated.

plaindocs:
I believe that comment covers it. Happy to try to help if there are questions. 

It would make it easier for other folks to help if we used a [file based Netlify config](https://docs.netlify.com/build/configure-builds/file-based-configuration/) instead of the Netlify UI.

MarcosBorgesPhD:
> @JoelMarcey -- could you help us out with the Netlify update? 🙏

In this PR, we removed the shebang from `make.py` so it can be invoked consistently across OS. Netlify now needs to call it through `uv` with `--frozen` to avoid modifying `uv.lock` (Issue #387).

Please replace:

`./make.py`

with:

`uv run --frozen make.py`

JoelMarcey:
@MarcosBorgesPhD I went ahead and made that update in Netlify if you want to give it a go again.

plaindocs:
Thanks @JoelMarcey !

@MarcosBorgesPhD looks like that failure is just a spec log again, so getting that merged (feel free to tag me for review and I'll get it in) should get this green.

MarcosBorgesPhD:
Thanks @JoelMarcey, @PLeVasseur, and @plaindocs!

> looks like that failure is just a spec log again ...

Building the habit! I’ll make sure to check for FLS updates before any PR next time 😉

plaindocs:
We did some work a while back on automatically updating the spec lock *if no guidelines* were affected (not sure about the state of that), but if you're working on a branch an not updating from from main regularly that probably won't help anyway.

MarcosBorgesPhD:
> We did some work a while back on automatically updating the spec lock _if no guidelines_ were affected ...

I’ll check our automation when I get a chance! I saw something in the **reviewer-bot**, but I’m not sure about the triggers yet. I’ll work on it once we’ve sorted out the spec updates workflow.

plaindocs (approved):
As far as I can tell this looks good to go.

Issue #381, opened 2026-01-29
[CI / Automation] Only run PR tests that are relevant for the files being changed


felix91gr:
As can be seen in https://github.com/rustfoundation/safety-critical-rust-coding-guidelines/pull/216, our CI process runs a full test of the repo even for changes that are completely disjoint from things like the FLS `spec.lock` file. We should avoid doing this, since it's:

- A waste of computing resources, and
- An unnecessary blocker for PRs that have nothing to do with whatever may fail in CI

One could argue this is also a fault of `git` for not understanding that `README.md` and `spec.lock` are independent files, which is true: both Pijul and Darcs, which are patch-based VC systems, model this independence in their representation of the repo.

That being said, we're gonna be stuck with `git` for the time being, so we'll have to fix this at the CI level - which is where we can.

MarcosBorgesPhD:
Since this is impacting everything we push, I’ll be working on this today.

I’m starting with `build-guidelines.yml` since it handles the core build stages. 😉

```shell
Stage  Job name                  Workflow name       Workflow file           Events
0      check_rust_examples       Build               build-guidelines.yml    push,merge_group,pull_request,workflow_call
0      check_typos               Build               build-guidelines.yml    push,merge_group,pull_request,workflow_call
0      reviewer-bot-coverage     Reviewer Bot Tests  reviewer-bot-tests.yml  push,pull_request,workflow_dispatch
0      reviewer-bot-unit         Reviewer Bot Tests  reviewer-bot-tests.yml  push,pull_request,workflow_dispatch
0      reviewer-bot-integration  Reviewer Bot Tests  reviewer-bot-tests.yml  push,pull_request,workflow_dispatch
0      reviewer-bot-contract     Reviewer Bot Tests  reviewer-bot-tests.yml  push,pull_request,workflow_dispatch
0      lint-and-test             Ruff Checks         ruff-lint.yml           push,pull_request
1      build                     Build               build-guidelines.yml    push,merge_group,pull_request,workflow_call
```

There is room for a lot of improvement in these workflows, so any suggestions or feedback are more than welcome!

MarcosBorgesPhD:
👋 Help needed here!

I’m working on a new version of `build-guidelines.yml` with the following jobs:

## Job 1. `detect_changes`: [FAST]

Checks for file changes that require Rust example validation. If a specific file or anything inside a mapped directory matches, it returns `true`.

[...]

PR #1219, opened 2026-07-16
Improve test and build workflow with conditional execution (#381)

MarcosBorgesPhD:
This PR improves the `build-guidelines.yml` workflow by avoiding unnecessary tests when they are not needed. This is a first step toward addressing Issue #381.

The goal is not just to save compute time. It is also to avoid running unnecessary jobs that can slow down or block contributors because of unrelated issues elsewhere in the codebase.

## What changed

A `gatekeeper` job now determines whether the checks need to run.

When relevant files change:

- `check-rust-examples` runs
- `build` runs

Otherwise, those jobs are skipped.

> [!WARNING]
> We need to check whether the GitHub branch protection settings accept `build` reporting `success`, `failure`, and now `skipped`.

The `gatekeeper` was chosen instead of `paths` filters to keep the `build` status always reported for PRs. With `paths`, the workflow could be skipped entirely, leaving GitHub without a status to evaluate for merge/branch protection.

The `push` tag trigger was removed since tag builds are already covered by `workflow_call` from `deploy.yml`.

A `force_full_build` input (default: `true`) was added to `workflow_call` allowing `deploy.yml` and `nightly.yml` to bypass the gatekeeper and force a full build when needed.

## Files that trigger a full build

The workflow currently watches the following paths:

* `src/` (kept intentionally broad to avoid edge cases)
* `builder/`
* `exts/`
* `tests/rust-examples/`
* `scripts/extract_rust_examples.py`
* `scripts/migrate_rust_examples.py`
* `.github/workflows/build-guidelines.yml`
* `.github/workflows/check-rust-examples.yml`
* `make.py`
* `spec.lock`
* `pyproject.toml`
* `uv.lock`

There are several dependencies between the build scripts and the documentation, so I’d rather err on the side of running the build than accidentally missing a case. If we discover additional dependencies later, we can easily expand the watched paths.

I’ll keep monitoring the workflow after this lands to make sure there aren’t any unexpected side effects. 😅

### DIFF
--- a/.github/workflows/build-guidelines.yml
+++ b/.github/workflows/build-guidelines.yml
@@ -19,13 +19,89 @@ on:
         description: Build from src/spec.lock without retrieving the live FLS
         type: boolean
         default: false
+      force_full_build:
+        description: Force a full build for workflow calls, bypassing the gatekeeper
+        type: boolean
+        default: true
 
 jobs:
+  # Determine when a full build should run
+  gatekeeper:
+    runs-on: ubuntu-latest
+    outputs:
+      run_all: ${{ steps.detect-changes.outputs.run_all }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Detect changes that should trigger a full build
+        id: detect-changes
+        shell: bash
+        run: |
+          if [[ "${{ inputs.force_full_build }}" == "true" ]]; then
+            echo "🔥  Forcing full build on workflow_call."
+            echo "run_all=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          if [[ "${{ github.ref_name }}" = "main" ]]; then
+            echo "🔥  Forcing full build on push/merge to main."
+            echo "run_all=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          # Determine the base commit SHA to compare changes against
+          # Supporting pull request, merge, and push
+          BASE_REF="${{
+            github.event.pull_request.base.sha ||
+            github.event.merge_group.base_sha ||
+            github.event.before
+          }}"
+
+          # Fallback for new branch, manual run, or new tag
+          if [[ -z "$BASE_REF" ]] || [[ "$BASE_REF" =~ ^0+$ ]]; then
+            echo "🔥  Forcing full build for new branch, manual run, or new tag"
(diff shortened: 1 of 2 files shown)


MarcosBorgesPhD:
@felix91gr @plaindocs @PLeVasseur, could you assist me with this PR?

The building is failing now because the FLS was updated (#1236), which affects one guideline 😬. I'm not sure how to approach this, but I'll check in next week if I'm able to update the FLS.

PLeVasseur:
Hi @MarcosBorgesPhD -- I meant to get to this today, but please see my comment here:
https://github.com/Safety-Critical-Rust-Consortium/safety-critical-rust-coding-guidelines/issues/1237#issuecomment-5419691993

I want to get this unblocked. I will try to make that happen this week.

PLeVasseur:
Hi @MarcosBorgesPhD, #1239 has merged and should remove the FLS drift blocker you encountered. It also fixes the Netlify preview failure by pinning uv through repository-controlled configuration.
                                                                                                                                 
Could you rebase #1219 onto current main? It will conflict in .github/workflows/build-guidelines.yml because #1239 added FLS audit tests, release/offline inputs, required-check aggregation, build exit propagation, and FLS notices.

Your gatekeeper approach and the goal of #381 are still valid. During conflict resolution, please preserve the current main build behavior and layer the gatekeeper around it, keeping the required build context always reported. The new FLS tests and workflows should also be included in the relevant-path decision.

I’m happy to help work through the workflow conflict if useful. Thank you for sticking with this.

MarcosBorgesPhD:
Hi @PLeVasseur,

I rebased the workflow changes on the current main and reconciled them with #1239, so:

* #1239 makes the existing build behavior safer by adding FLS audit tests, required-check aggregation, and non-blocking FLS drift reporting.
* #381 avoids running that build pipeline when a PR does not affect it.

I put `fls_audit_tests` behind the same gatekeeper as `check_rust_examples` and `build`. This means an unrelated PR does not run the Rust example checks, FLS audit suite, or documentation build. `check_typos` remains independent.

The relevant-path list now also includes the new FLS audit test directories and keeps `src/` intentionally broad to avoid missing dependencies.

The required `build` check is still preserved for relevant builds and uses `always()` to aggregate prerequisite failures, so a failed Rust or FLS audit job cannot be hidden by dependency skipping.

I believe this preserves the current main behavior from #1239 while applying the optimization originally intended by #381.

Happy to contribute with this optimization!

plaindocs:
This is exciting! 🎉

PR #1230, opened 2026-07-31
Add draft guideline for avoiding unused macros

plaindocs:
This is a super simple guideline that states that you shouldn't define macros you don't use.

I'm also using it to test the [local workflow](https://github.com/Safety-Critical-Rust-Consortium/safety-critical-rust-coding-guidelines/blob/main/README.md), will open an issue after.

### DIFF
--- a/src/coding-guidelines/macros/gui_GQm4lfJ4zYUD.rst
+++ b/src/coding-guidelines/macros/gui_GQm4lfJ4zYUD.rst
@@ -49,7 +49,7 @@ Avoid unused macros
             }
 
             fn main() {
-              self::used_macro!()
+              self::used_macro!{}
             }
 
 .. bibliography::
--- a/src/coding-guidelines/macros/gui_GQm4lfJ4zYUD.rst
+++ b/src/coding-guidelines/macros/gui_GQm4lfJ4zYUD.rst
@@ -28,6 +28,7 @@ Avoid unused macros
         .. rust-example::
             :no_run:
 
+            #[allow(unused_macros)]
             macro_rules! never_used {
                 () => {};
             }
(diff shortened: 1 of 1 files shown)


src/coding-guidelines/macros/gui_GQm4lfJ4zYUD.rst
```diff
@@ -0,0 +1,71 @@
...
+        .. rust-example::
+
+            macro_rules! used_macro {
+            ($t:expr) => {
+                println!("MACRO TEXT: {}", $t);
+                }
+            }
+
+            fn main() {
+            used_macro!("I am used, therefore I am.");
```
danieldia-dev: For aesthetic reasons, perhaps indent the `used_macro!(...)`?

danieldia-dev:
Hi @plaindocs!

Following through with the checklist: 
- [x] Guideline title is clear and follows conventions
- [x] Amplification section expands on the title appropriately
- [x] Rationale explains the "why" effectively
- [x] Non-compliant example clearly shows the problem
- [x] Compliant example clearly shows the solution
- [x] Code examples compile
- [x] FLS paragraph ID is correct

I think all is good, but regarding this guideline in particular, I'd suggest making two separate guidelines (mimicking the way this is done in `rustc`): an equivalent of `unused_macros` (i.e. an entire `macro_rules!` definition is never invoked anywhere) and an equivalent of `unused_macro_rules` (i.e. individual match arms or rules inside the `macro_rules!` block never matches any invocation). Let me know what you think, as this may turn out to be a redundant way to go about this.

danieldia-dev (approved):

danieldia-dev (approved):

